PR #30, opened 2017-12-26
inertia deploy [REMOTE] (up, down, status, reset) and authentication

bobheadxi:
:hourglass: **Status**: WIP, ready for review

:tickets: **Ticket(s)**: Closes #27, closes #32, closes #33, closes #38

---

## :construction_worker: Changes
- `inertia deploy [REMOTE] down`:
    - Checks if inertia is only active container, errors if it is
    - Kills all other containers

- `inertia deploy [REMOTE] up`:
    - If no git repository on server, clone using the repo URL in the `post` request
    - `deploy()` project

- `inertia deploy [REMOTE] status`:
    - Returns a list of active non-daemon containers

- `inertia deploy [REMOTE] reset`:
    - kills containers and deletes contents of project folder

- authentication using github token (#39)

- Push events to Daemon:
    - If no git repository on server, clone using the repo URL in the event
    - `deploy()` project

- `deploy()`:
    - Pulls from working tree, fresh clone if that fails (go-git doesn't support force pull)
    - Run `docker-compose` using similar docker-in-docker workaround to `daemon-up`

- A bunch of minor typo fixes, refactored some shared helper functions to `util.go`

## Todo
- [x] Clone on `up`
- [x] Error codes and better responses
- [x] `docker-compose`
- [x] Authorization 🤔 (see #38 )
- [x] Tests (minimal 😢 )

## :flashlight: Testing Instructions
```bash
inertia init
inertia remote add local 0.0.0.0 -u robertlin
docker build -t inertia .
```
This next thing basically `bootstrap`s without pulling master inertia:
```bash
sudo docker run -p 8081:8081 -e HOME -v /var/run/docker.sock:/var/run/docker.sock -v /usr/bin/docker:/usr/bin/docker -v $HOME:/app/host -e SSH_KNOWN_HOSTS='/app/host/.ssh/known_hosts' -e HOME=$HOME --name="inertia-daemon" inertia
```
If needed, do `docker rm -f $(docker ps -a -q)`
Now you can:
```bash
inertia remote status local
inertia deploy local up
inertia deploy local down
inertia deploy local status
```

For example, with [Sleuth](https://github.com/ubclaunchpad/sleuth):
```bash
Roberts-MacBook-Pro:inertia robertlin$ inertia deploy local up
(Status code 201) Project up

Roberts-MacBook-Pro:inertia robertlin$ inertia deploy local status
(Status code 200) 7b7be0b7097a26169e17037f4220fd0ce039bde1 refs/heads/master
Active containers:
[...]

PR #33, opened 2017-12-27
Add repo to post request

chadlagore:
:hourglass: **Status**: WIP

:tickets: **Ticket(s)**: #42

---

+ Added the repo name to the `/up` request
+ We punted on this because it might show up via webhook when you add the deploy key - maybe not - easy enough to add it here.
+ SUPER ALPHA. Only knows `origin` :joy: just use this branch for testing.


chadlagore:
**Handled in #30**

Issue #38, opened 2018-01-03
Authorization


chadlagore:
## Add some auth to the daemon

One promising approach is to use a [JWT](https://jwt.io/). We've already generated an RSA key server-side. We can use it to [create a JWT](https://github.com/dgrijalva/jwt-go) during the server bootstrapping process, and securely pipe it back to the client over SSH for storage in it's config folder. All of this should happen without involving the user.

PR #39, opened 2018-01-04
Chad/#38 authorization

chadlagore:
:hourglass: **Status**: _Ready for review_

:tickets: **Ticket(s)**: #38 

---

+ Added an `authorization` function decorator.
+ Added token generation functionality, passed securely over SSH.
+ Included comments on how to apply it to endpoints.
+ Implemented a trivial endpoint for health checks.
+ Added some testing around auth.

```bash
go test ./cmd -cover
```


cmd/util.go
```diff
@@ -84,3 +93,61 @@ func removeContents(directory string) error {
...
+	key, err := getGithubKey()
+	if err != nil {
+		return nil, err
+	}
+	return []byte(key.String()), nil
+}
+
+// getGithubKey returns the key generated by 'inertia remote bootstrap [REMOTE]'
+func getGithubKey() (ssh.AuthMethod, error) {
+	pemFile := "/app/host/.ssh/id_rsa_inertia_deploy"
```
bobheadxi: A bit hesitant about putting this function in `util` since that path (`/app/host/...`) applies specifically to the daemon container

chadlagore: Mmmm. How about we make this function more general? `getKeyFromPem(pemFile string)`. Then we could possibly unit test it too.

chadlagore: Actually I think [this](https://github.com/ubclaunchpad/inertia/blob/master/cmd/init.go#L212) is the "go" way to write testable code that does file IO. Something like `func getKeyFromPem(pemFile io.Reader)`

chadlagore: I can move it back for the time being :+1: Something to think about.

cmd/util.go
```diff
@@ -84,3 +93,61 @@ func removeContents(directory string) error {
...
+// sending HTTP requests to the daemon server.
+func generateToken(key []byte) (string, error) {
+	// No claims for now.
+	return jwt.New(jwt.SigningMethodHS256).SignedString(key)
+}
+
+// getAPIPrivateKey returns the private RSA key to authenticate HTTP
+// requests sent to the daemon. For now, we simply use the GitHub
+// deploy key.
+func getAPIPrivateKey(*jwt.Token) (interface{}, error) {
```
bobheadxi: @chadlagore why does this return `interface{}`?

Issue #46, opened 2018-01-06
Restructure project


bobheadxi:
Possible packages:
- `cmd`: top-level package for cobra and command line stuff (probably excluding daemon commands?)
- `daemon`: server-side functionality (daemon)
- `client`: client-side functionality (init, remote, deploy, bootstrap)
- `common`: bottom-level package for shared types, utils, bootstrap scripts, etc

```
        cmd
      /  |  \
 daemon  |  client
      \  |  /
       common
```

Issue #35, opened 2017-12-31
Inertia versioning


bobheadxi:
Might be a good idea to version inertia so that the desired version of the tool can be used, such as:
```
inertia init -v 1.0
```
Because `bootstrap` pulling the latest version of inertia could cause problems with version discrepancies

Bonus for being able to specify a branch version of inertia for testing 😛 

As noted in #95, also add to Travis builds

bobheadxi:
After discussion:
> `inertia [REMOTE] init` will always deploy the tag `latest` daemon. `inertia [REMOTE] init -v <tag or hash>` where tag or hash can be an image tag or hash on the dockerhub image repo.

bobheadxi:
Not urgent - moving back to `v0.0.3`

PR #52, opened 2018-01-09
Split project into 4 packages, add Makefile

bobheadxi:
:hourglass: **Status**: WIP, but everything seems to work so far

:tickets: **Ticket(s)**: #46

---
## Motivation
- client side code has little overlap with daemon code (eg no need for docker) and vice versa (daemon has no access to local config files, should not be using bootstrap, etc)
- what functions do and where variables come from is more clear: 
```
client.GetConfigFile()
daemon.Run()
daemon.DefaultPort
```
etc
- CLI definitions might be cleaner in a separate package

- Inertia is now 4 packages as outlined in #46 

- Makefile for the lazy

```
make test
make docker
```
and test on gcloud


Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+PACKAGES = `go list ./... | grep -v vendor/`
+
+all: inertia
+
+inertia:
+	go build
+
+test:
+	go test $(PACKAGES) --cover
```
chadlagore: Might as well add -race

Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+test-verbose:
+	go test $(PACKAGES) -v --cover
+
+clean: inertia
+	rm -f inertia
+	rm -f cover.html
+
+docker:
```
chadlagore: I thought we’d eventually make this an automated build, but this works for development

Makefile
```diff
@@ -0,0 +1,25 @@
...
+	docker build -t ubclaunchpad/inertia .
+	docker push ubclaunchpad/inertia
+
+bootstrap:
+	go-bindata -o client/bootstrap.go client/bootstrap/...
```
chadlagore: +1. we should also either make the bootstrap file either a part of the main package or use a python script to replace the package header inside the bootstrap file after we automatically generate it. I’m partial to the former.

Makefile
bobheadxi: TIL gotest has a race condition detector 😱 

Speaking of which need to make sure daemon doesn't do bad things when a bunch up people ping it to do things at the same time

Makefile
bobheadxi: Could change this to tag it as a test build in the future and add some `inertia [REMOTE] init` flags (ie `test-build`) to pull that build? Not sure if dockerhub supports that

Makefile
bobheadxi: I don't think we can put it in `main` since that would be a cyclic dependency if `client` uses it right?

Makefile
chadlagore: These two things happen at very different layers. One is when we’re tweaking daemon functionality, and one is when the user is trying to deploy the application. I suppose I could see creating a staging image for testing inertia that receives arbitrary pushes, but I’m not too in love with adding this type of functionality to the CLI...

Makefile
chadlagore: Yeah exactly, I think that’s why I didn’t do it in the first place 🤔 maybe there’s a flag we can use to set the package name?

Makefile
bobheadxi: Actually I meant something closer to #35, where we implement something that keeps the daemon version on `[REMOTE] init` in sync with the local inertia, except everything that's not a proper tagged release we just call it `snapshot` maybe

So:
`inertia init -v 0.0.1` would download the docker build tagged as `v0.0.1`
`inertia init -v snapshot` would download the docker build tagged as `SNAPSHOT` which is dangerous and could be anything but I think that's the point of snapshot builds?

Makefile
bobheadxi: It's [somewhere](https://github.com/jteeuwen/go-bindata/blob/master/config.go#L28)

Issue #59, opened 2018-01-14
Secure client-server communication


jordanschalm:
Right now the daemon sets up an HTTP (not HTTPS) listener. This makes the token authorization system vulnerable to replay attacks since the token is sent in the clear.

To fix: Use HTTPS exclusively. Maybe there's a way to generate a cert when creating a new remote, and informing the client-side to trust that cert?

chadlagore:
+1 on this, not done in the first place out of pure ignorance. Must do :+1:.

jordanschalm:
**References:**
* [generating certs](https://golang.org/src/crypto/tls/generate_cert.go)
* [listening on TLS](https://golang.org/pkg/net/http/#ListenAndServeTLS)

bobheadxi:
More references:
- [automated certificates](https://godoc.org/golang.org/x/crypto/acme/autocert)
- [serving](https://blog.kowalczyk.info/article/Jl3G/https-for-free-in-go-with-little-help-of-lets-encrypt.html)

PR #63, opened 2018-01-19
POC SSL Certificate

chadlagore:
:hourglass: **Status**: WIP

:tickets: **Ticket(s)**: #59

---

```bash
./gen-cert.sh  # lots of input - ideally pipe this script over ssh and fill in with defaults
sudo go run main.go  # port 443 requires sudo, bruno will be pissed
```

Elsewhere...

```bash
$ curl -sL https://localhost:443
```

Produces server-side error:

```bash
2018/01/18 21:00:43 http: TLS handshake error from [::1]:53606: remote error: tls: unknown certificate authority
```

@jordanschalm and I discuss:

> ... because we control both the client and the server, and we have to ssh pipe, we can just provide any identity that both sides can agree on. We don’t actually care about identities here, all we want is the encryption, so we can kind of fudge the identity part.

Discuss?


jordanschalm:
@bfbachmann I don't think that's a concern. Since the bootstrapping connection occurs over a secure channel we can inform the client of what cert the server has generated, so the client knows what it should trust. 

The cert *does* come from a trusted authority (our code, running on the server), but the library doesn't know that without some extra work on our part. That's the hard part.

poc/tls/gen-cert.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+openssl genrsa -out server.key 2048
```
jordanschalm: Can we remove the dependency on openssl?

poc/tls/main.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+import (
+	"log"
+	"net/http"
+)
+
+func HelloServer(w http.ResponseWriter, req *http.Request) {
```
jordanschalm: Hahaha damn this is simple 😆

chadlagore:
May be of use https://godoc.org/github.com/spacemonkeygo/openssl

poc/tls/gen-cert.sh
chadlagore: May be of use https://godoc.org/github.com/spacemonkeygo/openssl

chadlagore:
No @bobheadxi, this will be handled over SSH prior to booting the daemon container 😄

PR #64, opened 2018-01-22
'inertia remote add' walkthrough

### LINKED ISSUE #53: Remote Add Walkthrough
We currently silently expect certain parameters on `inertia remote add`. When the user doesnt actually give them, we use defaults which are not good.

Change it to this,

```
$ inertia add remote instance-1
Enter location of pem file:
Enter IP:
```

Then try to extract the SSH user from the pem public file (this has the benefit of failing if we cant find the public key - which would be necessary to authenticate over SSH anyways).

bobheadxi:
:hourglass: **Status**: redy redy

:tickets: **Ticket(s)**: Close #53 'inertia remote add' walkthrough

---

- Walkthrough for the 'remote add' command
- Left `port` as a flag... but maybe should make it part of the walkthrough? 🤔 

```
go install
inertia init
inertia remote add gcloud
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,16 @@ $> inertia init
 
 $> inertia remote add glcoud 35.227.171.49 -u root -i /path/to/my/.ssh/id_rsa
 Remote 'glcoud' added.
+$> inertia remote add gcloud
+Enter location of PEM file (leave blank to use '/Users/yourspecialname/.ssh/id_rsa'):
+/path/to/my/.ssh/id_rsa
+Enter IP address of remote:
+35.227.171.49
+Enter user:
+root
+Remote 'gcloud' added.
+You can now run 'inertia gcloud init' to set this remote
+up for continuous deployment.
 
 $> inertia gcloud init
 Bootstrapping remote
--- a/README.md
+++ b/README.md
@@ -20,18 +20,19 @@ Inside of a git repository, run the following:
 ```bash
 $> inertia init
 
-$> inertia remote add glcoud 35.227.171.49 -u root -i /path/to/my/.ssh/id_rsa
-Remote 'glcoud' added.
 $> inertia remote add gcloud
 Enter location of PEM file (leave blank to use '/Users/yourspecialname/.ssh/id_rsa'):
 /path/to/my/.ssh/id_rsa
-Enter IP address of remote:
-35.227.171.49
 Enter user:
 root
-Remote 'gcloud' added.
-You can now run 'inertia gcloud init' to set this remote
-up for continuous deployment.
+Enter IP address of remote:
+35.227.171.49
+Port 8081 will be used as the daemon port.
+Run this 'inertia remote add' with the -p flag to set a custom port.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -22,6 +22,16 @@ $> inertia init
 
 $> inertia remote add glcoud 35.227.171.49 -u root -i /path/to/my/.ssh/id_rsa
 Remote 'glcoud' added.
+$> inertia remote add gcloud
+Enter location of PEM file (leave blank to use '/Users/yourspecialname/.ssh/id_rsa'):
+/path/to/my/.ssh/id_rsa
+Enter IP address of remote:
+35.227.171.49
+Enter user:
+root
```
chadlagore: So the user can in principle be parsed from the `/Users/yourspecialname/.ssh/id_rsa.pub` file. And in reality, we need to do _some_ verification on whether the user in the public key matches this name. That helps us avoid this type of failure:

```bash
inertia gcloud init
[Permission denied: public key]
```

I bet one of these go libraries we already have as a dependency lets you look at the public key user for a private key 🤔

One nice way to fix:

```
Enter user (press Enter to use default user for this PEM <default_user_we_parsed>):
```

chadlagore:
Might be a good idea to at least ask if they're ok with `8081` as the daemon port.

README.md
bobheadxi: @chadlagore If we can read the user of the key from the key, should we even allow manual input of a user?

chadlagore: Its true. If they enter the wrong username, we have no way to recover. Ideally, they don't have too. Another way is they could configure their ssh root once, and we always go rummaging around in there for keys.

cmd/remote.go
```diff
@@ -62,22 +62,48 @@ inerta remote status gcloud`,
...
+		_, err = fmt.Scanln(&response)
+		if err != nil {
+			response = defaultSSHLoc
+		}
+		pemLoc := response
+
+		fmt.Printf("Enter IP address of remote: \n")
+		_, err = fmt.Scanln(&response)
+		if err != nil {
+			log.Fatal("That is not a valid IP address - please try again.")
```
jordanschalm: I don't think this is an accurate error message. This error means `Scanln` failed, not that the input wasn't a valid IP

cmd/remote.go
```diff
@@ -62,22 +62,48 @@ inerta remote status gcloud`,
...
+		}
+		address := response
+
+		fmt.Printf("Enter user: \n")
```
jordanschalm: Use `Println` and remove the `\n`?

cmd/remote.go
bobheadxi: An error gets thrown if don't input anything as well (hit 'enter' straight away) - plus anything that causes `Scanln` to fail would be an invalid input right? 😅

README.md
bobheadxi: Might leave this for another ticket

chadlagore: 👍

PR #71, opened 2018-02-04
Simple dockerized test environment

### LINKED ISSUE #23: Dockerize Testing Environment
A lot of our initialization code does some small file system changes. This may be something we want to dockerize in order to test properly. Discuss?

```
container1 (fake local, file system changes) <---- ssh pipe ----> container2 (fake remote, accepts ssh commands)
```

bobheadxi:
:hourglass: **Status**: readyyyy

:tickets: **Ticket(s)**: Closes #23 

---

- Set up a "mock" VPS instance locally!
- SSH key for testing connection with mock VPS
- Build specific version of Ubuntu or CentOS for your mock VPS
   - Ubuntu 14.04+ working fine
   - CentOS 7+ working with some hacks - need to test on a real VPS. CentOS 6 doesn't work
- Travis builds for different various VPS platforms and versions
- CLI: Support for custom SSH port
- CLI: Fixed err printing in SSH run

```bash
$> go install
$> make testenv-ubuntu 
# note the location of the key that is printed
$> cd /path/to/my/dockercompose/project
$> inertia init
$> inertia remote add local 
# PEM file: inertia test key, User: 'root', Address: 0.0.0.0 (standard SSH port)
$> inertia local init
$> inertia remote status local
Remote instance 'local' accepting requests at http://0.0.0.0:8081
```

You can also do things like:
```bash
$> make test VERSION=14.04 # runs tests against a Ubuntu 14.04 VPS
```

You can also build different platforms:
```bash
$> make test VPS_OS=centos VERSION=7
```

You can treat this container just as you would treat a real VPS. Example:
```go
func TestInstrumentedBootstrap(t *testing.T) {
	remote := getInstrumentedTestRemote()
	session := NewSSHRunner(remote)
	var writer bytes.Buffer
	err := remote.Bootstrap(session, "testvps", &Config{Writer: &writer})
	assert.Nil(t, err)

	// Check if daemon is online following bootstrap
	host := "http://" + remote.GetIPAndPort()
	resp, err := http.Get(host)
	assert.Nil(t, err)
	assert.Equal(t, resp.StatusCode, http.StatusOK)
	defer resp.Body.Close()
	_, err = ioutil.ReadAll(resp.Body)
	assert.Nil(t, err)
}
```


test_env/test_key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
chadlagore: This is late, but couldn't you use the Dockerfile to generate these keys? Then copy the private one out to the host?

```
RUN ssh-keygen <arguments>
```

```
docker cp test-env:/path/of/private/key /test_env
```

Just a thought, would keep them out of the repo.

bobheadxi: Hmmmm perhaps, but I think its a bit cleaner to always use the same key

client/bootstrap/docker.sh
```diff
@@ -39,3 +39,5 @@ if !(hash docker 2>/dev/null); then
         sh $DOCKER_DEST
     fi
 fi
+
+sudo service docker start
```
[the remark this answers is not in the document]
bobheadxi: lmao

PR #92, opened 2018-02-24
Define project name on docker-compose

### LINKED ISSUE #50: Set docker-compose project name
defaults to the folder name (in this case `project`) but [can be changed with a flag](https://github.com/docker/compose/issues/2982) - since it seems like some projects rely on hardcoded container names, would be good to pull project name from remote url just for `deploy()` so that our docker-compose doesn't append `project_` to the name of every container

PiggySpeed:
:hourglass: **Status**: WIP

:tickets: **Ticket(s)**: Closes #50

---

Sets the docker-compose project name to the repo name, rather than having `project_` prepended to the names of each container.

Currently looking for help with testing this change @bobheadxi


daemon/docker.go
```diff
@@ -83,13 +84,23 @@ func deploy(repo *git.Repository, cli *docker.Client) error {
...
+
+	remotes, err := repo.Remotes()
+	if err != nil {
+		return err
+	}
+
 	ctx := context.Background()
+	url := remotes[0].Config().URLs[0]
+	r, _ := regexp.Compile("(?:/)([^/]+)(?:.git)$")
+	repoName := r.FindStringSubmatch(url)[1]
```
bobheadxi: If you want to make these lines testable, one way would to be turn it into a helper function in package `common`:

```go
func getRepoNameFromURL(url string) string {
    // ...
}
```

Then you can just call that function alone for tests

daemon/docker.go
```diff
@@ -83,13 +84,23 @@ func deploy(repo *git.Repository, cli *docker.Client) error {
...
+
 	resp, err := cli.ContainerCreate(
 		ctx, &container.Config{
 			Image:      dockerCompose,
 			WorkingDir: "/build/project",
 			Env:        []string{"HOME:/build"},
-			Cmd:        []string{"up", "--build"},
+			Cmd:        []string{"-p", repoName, "up", "--build"},
```
bobheadxi: Nice!

daemon/docker.go
PiggySpeed: That sounds good, I'll try that out. 

Is there a way to test manually test this to make sure that a container is successfully created with the appropriate name? I know that the individual `-p` command works as expected for docker-compose, but I wanted to see if it works in the context of the entire method.

bobheadxi: Yep! Start with this: https://github.com/ubclaunchpad/inertia#testing

Then try doing:
```bash
docker ps
docker exec -it testvps sh
# this will take you into a shell inside the testenv container
docker ps
# should list containers that were started by daemon - these containers should have names prefixed by the project name you extracted
```

PiggySpeed: Thanks! It looks like it's working perfectly.

PiggySpeed: ... spoke too soon, it's not working >_<

common/util_test.go
```diff
@@ -12,13 +12,37 @@ import (
...
 
 var (
 	testPrivateKey = []byte("very_sekrit_key")
 	testToken      = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.AqFWnFeY9B8jj7-l3z0a9iaZdwIca7xhUF3fuaJjU90"
 )
 
+func getMockRepo(url string) (*git.Repository, error){
+	memory := memory.NewStorage()
+
+	if url[len(url) - len(".git"):] != ".git" {
```
bobheadxi: Since this is a function purely for testing, I don't think there's really a need to provide `url` as a parameter - I think it should be hardcoded. This would eliminate the need for this URL check and make the behaviour of this test helper more predictable. Otherwise this is pretty clean!

common/util_test.go
```diff
@@ -99,3 +123,17 @@ func TestFlushRoutine(t *testing.T) {
...
+
+func TestGetProjectName(t *testing.T) {
+	testRemote1, err := getMockRepo("https://github.com/ubclaunchpad/inertia.git")
+	assert.Nil(t, err)
+	repoName, err := GetProjectName(testRemote1)
+	assert.Equal(t, "inertia", repoName)
+
+	a, err := getMockRepo("https://github.com/ubclaunchpad/inertia")
+	if err != nil {
+		fmt.Println(err)
```
bobheadxi: Don't print in tests, since differences aren't automatically caught - use one of the `assert` functions 👍

daemon/docker.go
```diff
@@ -83,11 +83,19 @@ func deploy(repo *git.Repository, cli *docker.Client, out io.Writer) error {
 	// See https://cloud.google.com/community/tutorials/docker-compose-on-container-optimized-os
 	fmt.Fprintln(out, "Building project...")
 	ctx := context.Background()
+
+	repoName, err := common.GetProjectName(repo)
+	if err != nil {
+		fmt.Println(repoName)
```
bobheadxi: Not sure if this is just for work in progress print testing but when printing in the `daemon`, notice that almost every function now has an `io.Writer` parameter. `io.Writer` is an interface that is taken as a parameter by many of Go's functions to point towards a particular output. Interfaces in Go are implicitly inherited so, in this case, any class that implements a function `Write()` is an `io.Writer`!  😮 

On line 84 of this file:
```go
// print to output "out"
fmt.Fprintln(out, "Building project...")
```

daemon/docker.go
```diff
@@ -83,11 +83,19 @@ func deploy(repo *git.Repository, cli *docker.Client, out io.Writer) error {
...
+		return err
+	}
+
 	resp, err := cli.ContainerCreate(
 		ctx, &container.Config{
 			Image:      dockerCompose,
 			WorkingDir: "/build/project",
 			Cmd: []string{
+				"-p", repoName,
```
bobheadxi: Try updating this whole section:
```go
	resp, err := cli.ContainerCreate(
		ctx, &container.Config{
			Image:      dockerCompose,
			WorkingDir: "/build/project",
			Env:        []string{"HOME=/build"},
			Cmd: []string{
				"up",
				"--build",
			},
		},
		&container.HostConfig{
			Binds: []string{
				"/var/run/docker.sock:/var/run/docker.sock",
				os.Getenv("HOME") + ":/build",
			},
		}, nil, "docker-compose",
	)
```

Unfortunately I think I broke it a little while back 😓

daemon/docker.go
PiggySpeed: Oh, sorry! This was from debugging, I'll remove it.

bobheadxi: Still a useful trick - if you print to `out`, when you test it out you can just use `inertia local logs --stream` and the output will be sent to you 😄

common/util_test.go
PiggySpeed: Agreed, it would make more sense to have the parameter when the function is reused.

Done.

common/util_test.go
PiggySpeed: Thanks,

Done.

daemon/docker.go
PiggySpeed: Done.

bobheadxi:
Megacheck is reporting this: `common/util_test.go:105:12: this value of err is never used (SA4006)`

Do you mind giving that a quick fix? Thanks!

common/util_test.go
```diff
@@ -74,3 +98,17 @@ func TestFlushRoutine(t *testing.T) {
 		i++
 	}
 }
+
+func TestGetProjectName(t *testing.T) {
```
bobheadxi: I suppose this probably isn't needed anymore, right? 😅

daemon/docker.go
```diff
@@ -57,12 +57,20 @@ func deploy(repo *git.Repository, branch string, cli *docker.Client, out io.Writ
...
+		return err
+	}
+
 	resp, err := cli.ContainerCreate(
 		ctx, &container.Config{
 			Image:      dockerCompose,
 			WorkingDir: "/build/project",
 			Env:        []string{"HOME=/build"},
 			Cmd: []string{
+				"-p", repoName,
```
bobheadxi: 🔥

client/deployment.go
```diff
@@ -20,6 +20,7 @@ type Deployment struct {
 	*RemoteVPS
 	Repository *git.Repository
 	Auth       string
+	Project    string
 }
 
 // DaemonRequester can make HTTP requests to the daemon.
```
bobheadxi: Do you mind completely removing this interface? it doesn't do much and is tricky to keep up to date. thanks!

client/deployment_test.go
```diff
@@ -65,6 +65,7 @@ func TestUp(t *testing.T) {
 		err = json.Unmarshal(body, &upReq)
 		assert.Nil(t, err)
 		assert.Equal(t, "myremote.git", upReq.GitOptions.RemoteURL)
+		assert.Equal(t, "test_project", upReq.Project)
```
bobheadxi: 🥇 nicee

deployment.go
```diff
@@ -30,7 +30,7 @@ var deploymentUpCmd = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
-		resp, err := deployment.Up(stream)
+		resp, err := deployment.Up(deployment.Project, stream)
```
bobheadxi: 🔥 🔥 🔥

daemon/github.go
```diff
@@ -83,7 +83,7 @@ func processPushEvent(event *github.PushEvent) {
 			return
 		}
 		defer cli.Close()
-		err = deploy(localRepo, branch, cli, os.Stdout)
+		err = deploy(localRepo, branch, *repo.Name, cli, os.Stdout)
```
PiggySpeed: @bobheadxi 

What do you think about using `repo.Name` here? In the other call to `deploy()`, we use the project folder name e.g. `inertia-deploy-test`, but we wouldn't expect them to differ?

bobheadxi: Ah, forgot about this one.

What I might do is save `projectName` as a variable and always use that - I think that should be safe, since automated deployment can not happen until we receive an `up` request, which will include the `project` in its payload.

common/util_test.go
PiggySpeed: yup! removed

client/deployment.go
PiggySpeed: done

PiggySpeed: Actually @bobheadxi, it turns out to be important for the current solution; I'll try to find other ways that don't rely on this interface.

deployment.go
PiggySpeed: 🔥 🔥 🔥 🔥

PiggySpeed:
To be honest though, I feel a bit bad about creating a global variable.

bobheadxi:
If you merge `master` into your branch, you'll find that project deployment now has its own `daemon/inertia/project` package, which will limit the scope of your variable a little bit! I don't think it is too bad.

Alternatively, a daemon-side `Project` class could be made that contains project parameters and used for deployment, but at this point I'm not sure it is worth the effort

daemon/docker.go
```diff
@@ -17,9 +17,15 @@ import (
 	git "gopkg.in/src-d/go-git.v4"
 )
 
+var projectName = ""
```
bobheadxi: I feel that this could use a default of some sort (perhaps `"project"`), for no real reason other than this looks a bit strange 😅

PiggySpeed: done

PR #93, opened 2018-02-24
Deployment Logs

### LINKED ISSUE #37: Access deployment logs
```
inertia [REMOTE] logs -c [MY_CONTAINER_NAME]
```

Allow users to view/stream logs from their inertia deployment

bobheadxi:
:hourglass: **Status**: Ready

:tickets: **Ticket(s)**: Closes #37

---

- Stream Inertia daemon logs with `inertia local logs --stream`, or access most recent logs with `inertia local logs`
- Stream docker-compose logs or specific containers using `inertia local logs docker-compose` or `inertia local logs mycontainer` specifically
-Reworked logging, ~~though this is bugged right now:~~

~~https://github.com/ubclaunchpad/inertia/blob/rob/%2337-deployment-logs/daemon/handler.go#L196 doesn't seem to properly return the error to the client - I think its still floating around somewhere because if I run `inertia local up` again, the line that was supposed to be returned shows up in the container logs alongside the first log from the second call~~ 
FIXED as of 754d625, turns out a `newline` is needed for the flush to trigger

```bash
inertia local init
inertia local up
inertia local logs --stream # stream inertia daemon logs
inertia local logs docker-compose --stream
```

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -22,8 +22,9 @@ const (
 	DaemonOkResp = "I'm a little Webhook, short and stout!"
 )
 
-// UpRequest is the body of a up request to the daemon.
-type UpRequest struct {
-	Stream bool   `json:"stream"`
-	Repo   string `json:"repo"`
+// DaemonRequest is the configurable body of a request to the daemon.
+type DaemonRequest struct {
+	Stream    bool   `json:"stream"`
+	Repo      string `json:"repo,omitempty"`
+	Container string `json:"container,omitempty"`
 }
--- a/daemon/handler.go
+++ b/daemon/handler.go
@@ -15,6 +15,8 @@
 package daemon
 
 import (
+	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -24,6 +26,7 @@ import (
 	"strings"
 
 	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/docker/docker/api/types"
 	docker "github.com/docker/docker/client"
 	"github.com/google/go-github/github"
 	log "github.com/sirupsen/logrus"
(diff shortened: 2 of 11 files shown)


common/types.go
```diff
@@ -22,8 +22,9 @@ const (
...
 
-// UpRequest is the body of a up request to the daemon.
-type UpRequest struct {
-	Stream bool   `json:"stream"`
-	Repo   string `json:"repo"`
+// DaemonRequest is the configurable body of a request to the daemon.
+type DaemonRequest struct {
+	Stream    bool   `json:"stream"`
+	Repo      string `json:"repo,omitempty"`
+	Container string `json:"container,omitempty"`
```
chadlagore: As handy as this JSON language is, it feels like a smell to me

daemon/handler.go
```diff
@@ -48,6 +51,39 @@ const (
...
+		log.Println("Failed to pull docker-compose image - shutting down daemon.")
+		cli.Close()
+		return
+	}
+	cli.Close()
+
+	// Run daemon on port
+	log.Println("Serving daemon on port " + port)
+	mux := http.NewServeMux()
+	// Example usage of `authorized' decorator.
```
chadlagore: Can kill my old silly comments, think it’s clear what’s goin down here now😜

common/types.go
bobheadxi: hm what do you mean?

chadlagore: It just feels like it got hacked onto the language. Its very handy, but something feels strange about it. I'm just murmuring out loud, ignore me :+1:

bobheadxi: true, functionality-wise I think it's quite nice though 😛

Issue #72, opened 2018-02-06
Merge `inertia remote status` with `inertia [REMOTE] status`


bobheadxi:
Right now there are two remote-status-related commands:
```bash
$> inertia remote add gcloud
$> inertia remote status gcloud
Remote instance 'gcloud' accepting requests at http://myhost.com:8081
$> inertia gcloud status
(Status code 200) 7b7be0b7097a26169e17037f4220fd0ce039bde1 refs/heads/master
There are currently no active containers.
```

I'm in favour of putting the `inertia remote status` functionality into `inertia [REMOTE] status`:
```bash
$> inertia gcloud status
Could not connect to daemon
Try running inertia [REMOTE] init
$> inertia gcloud init
$> inertia gcloud status
(Status code 200) Daemon at remote 'gcloud' online at http://myhost.com:8081
Project status:
    7b7be0b7097a26169e17037f4220fd0ce039bde1 refs/heads/master
    There are currently no active containers.
```

PR #97, opened 2018-02-26
Migrate 'cmd' to 'main', licence update

### LINKED ISSUE #94: Move cmd package into main
Does it really need to be in a separate package? It's already at the top of the package food chain and the only thing `main` does is... call a function in `cmd`

My bad, tbt #46

### LINKED ISSUE #98: Standardise licences
Right now we have two licences going on:
- `Apache 2.0` in individual Go files
- `MIT` in our repository licence

paging @chadlagore

bobheadxi:
:tickets: **Ticket(s)**: 
- Closes #94 
- Closes #98 

---

This is largely a simple code migration. ⏫ 

* All command-line functionality now in 'main' package
* 'inertia daemon' commands are now completely disabled outside of the inertia docker image to prevent someone from inadvertently starting up a daemon locally or, even worse, name one of their remotes `daemon` and having it conflict with the command 😢 
* Removed all references to the `Apache 2.0` licence - we are already under `MIT`

```bash
go install
inertia daemon run # should no longer work
```


client/bootstrap/daemon-up.sh
```diff
@@ -33,7 +33,7 @@ sudo docker run -d --rm \
     -p "$PORT":8081 \
     -v /var/run/docker.sock:/var/run/docker.sock \
     -v $HOME:/app/host \
-    -e DAEMON="true" \
+    -e INERTIA_DAEMON='true' \
```
PiggySpeed: Just an observation - it looks like use of " " and ' ' is inconsistent in our shell scripts.

bobheadxi: That's a good point... I'm not actually sure how interchangeable they are 🤔 

Will look into it

bobheadxi:
Thanks for the review @PiggySpeed!

bobheadxi:
@rogermyang This one will affect your ticket (#72), remember to make sure you are up to date with `master` when I merge later 😄

client/bootstrap/daemon-up.sh
bobheadxi: I think the way we have it now is fine: [see this explanation for single vs double quotes in bash](https://stackoverflow.com/questions/6697753/difference-between-single-and-double-quotes-in-bash)

Single for string literals, double for things like `"$MYVAR"`

PiggySpeed: Cool, thanks for the link!

PR #125, opened 2018-03-18
Daemon repackaging

### LINKED ISSUE #103: Separate Daemon and CLI code
Right now our release binaries clock in at just over 15mb, which seems a bit large for what this application should be. Some platform binaries are even larger - take a look at the [Banana release](https://github.com/ubclaunchpad/inertia/releases/tag/v0.0.2)

We should be able to treat `daemon` as a package completely separate from the rest of the codebase (excluding `common`), but need to do more research. This would allow us to avoid large dependencies (Docker) in our CLI binaries (which doesn't use the Docker package)

bobheadxi:
:tickets: **Ticket(s)**: Closes #103

---

- Only marginally smaller binary sizes for CLI (oh well)
- No more conditional inclusion of daemon commands in the CLI - the daemon is now a completely separate package (`main` in `daemon/inertia`)
- Updated Dockerfile and scripts accordingly

### To Do
- [ ] Make this play nice with coveralls, coverage reporting is a little whack

```
make test
```


bobheadxi:
Thanks for the review @PiggySpeed !

PR #132, opened 2018-03-20
More daemon restructuring

bobheadxi:
:tickets: **Ticket(s)**: Continuation of #103 

---

Further divided daemon into packages for less confusion (I hope!)

- `auth`
- `project`

Let me know if this doesnt make sense 😅


daemon/inertia/auth/certificate.go
```diff
@@ -58,7 +57,8 @@ func pemBlockForKey(priv interface{}) (*pem.Block, error) {
 	}
 }
 
-func generateCertificate(certPath string, keyPath string, host string) error {
+// GenerateCertificate creates an SSL certificate for HTTPS use
+func GenerateCertificate(certPath, keyPath, host, ecdsaCurve string) error {
```
PiggySpeed: Changes not really needed, but I have a weird feeling about the empty string as one of the switch cases. It maps directly to an RSA key when the function params is called `ecdsa`. Moreover, the message specified in the default case gives the false impression that `""` would map to an ECDSA curve. Substituting `"RSA"` for `""` and changing the error message could be an option. It would also help express the programmer's intent when they use it in the code, e.g. `err := GenerateCertificate(dir+"test.cert", dir+"test.key", "0.0.0.0", "RSA")`

What do you think?

bobheadxi: That's actually a really good point, didn't think about the error reporting nothing!

I will update this - thanks for pointing it out!

PiggySpeed: Also consider modifying the error message? It currently implies that a valid ECDSA scheme is the only way to go.

bobheadxi: 😅

daemon/inertia/project/repository_test.go
```diff
@@ -82,7 +82,11 @@ func TestUpdateRepository(t *testing.T) {
 func TestCompareRemotes(t *testing.T) {
 	cwd, err := os.Getwd()
 	assert.Nil(t, err)
-	inertia, _ := path.Split(cwd)
+
+	// Traverse back down to base directory
+	inertia := path.Dir(cwd)
+	inertia = path.Dir(inertia)
+	inertia = path.Dir(inertia)
```
PiggySpeed: Is there a better way to do this? It seems a bit fragile when we assume the base directory is exactly 3 levels deep.

bobheadxi: It certainly is, which is why it broke previously 😅 I think it's fine though - it's a test, and the directory structure of inertia in any environment (barring a structural change like this) should stay the same.

That said this was a very lazy fix, I'll change it to something a bit more clear:

```go
func TestCompareRemotes(t *testing.T) {
	// Traverse back down to root directory of repository
	repo, err := git.PlainOpen("../../../")
```

Good catch again, been getting sloppy 🙅‍♂️

daemon/inertia/project/repository_test.go
```diff
@@ -80,15 +79,8 @@ func TestUpdateRepository(t *testing.T) {
...
-	assert.Nil(t, err)
-
-	// Traverse back down to base directory
-	inertia := path.Dir(cwd)
-	inertia = path.Dir(inertia)
-	inertia = path.Dir(inertia)
-
-	repo, err := git.PlainOpen(inertia)
+	// Traverse back down to root directory of repository
+	repo, err := git.PlainOpen("../../../")
```
PiggySpeed: Are we safe to assume that the repo will always be 3 levels deep from the cwd?

bobheadxi: Yep! `cwd` for tests is always the location of the test itself, and until the next restructure (hopefully not too soon again 😅 ) `daemon/inertia/project` will always be `daemon/inertia/project` 🎉

PiggySpeed: Ok, I suppose the location shouldn't change too often.

PR #135, opened 2018-03-21
Roger/#72 merge inertia remote status

### LINKED ISSUE #72: Merge `inertia remote status` with `inertia [REMOTE] status`

rogermyang:
:tickets: **Ticket(s)**: Closes #72 Merge `inertia remote status` with `inertia [REMOTE] status` 

---

merged statusCmd implementation in remote.go  into deploymentStatusCmd.


deployment.go
```diff
@@ -109,6 +110,43 @@ var deploymentStatusCmd = &cobra.Command{
 	Requires the Inertia daemon to be active on your remote - do this by 
 	running 'inertia [REMOTE] up'`,
 	Run: func(cmd *cobra.Command, args []string) {
+		config, err := client.GetProjectConfigFromDisk()
```
bobheadxi: I don't think we need most of this code to retrieve the remote - this is basically all already handled by:

```go
deployment, err := client.GetDeployment(strings.Split(cmd.Parent().Use, " ")[0])
// deployment has all the information for the remote
```

deployment.go
```diff
@@ -109,6 +110,43 @@ var deploymentStatusCmd = &cobra.Command{
...
+		remote, found := config.GetRemote(remotename)
+		if !found {
+			println("No such remote " + remotename)
+			println("Inertia currently supports one remote per repository")
+			println("Run `inertia remote -v' to see what remote is available")
+			os.Exit(1)
+		}
+
+		host := "http://" + remote.GetIPAndPort()
+		remoteResp, err := http.Get(host)
```
bobheadxi: We don't need this `http.Get` either. Have a look at the code for [`deployment.Status()` - notice it makes an HTTP request to the daemon already](https://github.com/ubclaunchpad/inertia/blob/master/client/deployment.go#L85). The following code is around [line 155](https://github.com/ubclaunchpad/inertia/pull/135/files#diff-4b475da4d1c32296f086ca8abdb29c83R155) of this revision:

```go
# makes an HTTP request to daemon
resp, err := deployment.Status()
if err != nil {
    # oh no! request failed
    log.WithError(err)
}
```

If this operation fails, we know the daemon is not online. If it succeeds, we know the daemon is online, and we can report that the daemon is listening at x address 🎉 

Speaking of which, there is a bug somewhere here that would be awesome if you could dig it up - the `status` command currently breaks with a null pointer error if the daemon is offline (#70) 😢 if you can find it you can close two tickets at once 🏋️

deployment.go
```diff
@@ -109,6 +110,43 @@ var deploymentStatusCmd = &cobra.Command{
...
+		if err != nil {
+			println("Could not connect to daemon")
+			println("Try running inertia [REMOTE] init")
+			return
+		}
+		defer remoteResp.Body.Close()
+		remoteBody, err := ioutil.ReadAll(remoteResp.Body)
```
bobheadxi: See my previous comments - all this has essentially been done already by [`deployment.Status()`](https://github.com/ubclaunchpad/inertia/blob/master/client/deployment.go#L85)

deployment.go
```diff
@@ -127,7 +165,8 @@ var deploymentStatusCmd = &cobra.Command{
 
 		switch resp.StatusCode {
 		case http.StatusOK:
-			fmt.Printf("(Status code %d) %s\n", resp.StatusCode, body)
+			fmt.Printf("(Status code %d) Daemon at remote '%s' online at %s\n", resp.StatusCode,remotename, host)
```
bobheadxi: Awesome! This line is really all we need 🥇

deployment.go
```diff
@@ -127,7 +165,8 @@ var deploymentStatusCmd = &cobra.Command{
...
+			fmt.Printf(" %s", body)
```
bobheadxi: Probably want to remove the space before `%s` here - might make formatting a bit weird

remote.go
```diff
@@ -4,8 +4,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
```
bobheadxi: 🚧 Let's not leave commented code around! If it isn't needed, just delete it 🚧

remote.go
```diff
@@ -263,7 +218,7 @@ func printRemoteDetails(remote *client.RemoteVPS) {
 func init() {
 	rootCmd.AddCommand(remoteCmd)
 	remoteCmd.AddCommand(addCmd)
-	remoteCmd.AddCommand(statusCmd)
+	//remoteCmd.AddCommand(statusCmd)
```
bobheadxi: Same here - if we don't need code anymore, just delete it. We can see the history in the git history, so don't worry about losing anything! 📖 

Leaving commented code, especially without an explanation, usually means no one will ever remove it (out of fear it might actually be important), and eventually it just gathers up all over the place 😢

deployment.go
```diff
@@ -10,8 +10,9 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-
+	"github.com/ubclaunchpad/inertia/common"
 	"github.com/ubclaunchpad/inertia/client"
+	"os"
```
bobheadxi: Are you using a Go linter with your IDE? For me at least, the Golang extension for [Visual Studio Code](https://code.visualstudio.com) will automatically sort these imports correctly and add/remove things as needed (the native `os` import should be above these third party imports)

deployment.go
```diff
@@ -109,6 +110,43 @@ var deploymentStatusCmd = &cobra.Command{
 	Requires the Inertia daemon to be active on your remote - do this by 
 	running 'inertia [REMOTE] up'`,
 	Run: func(cmd *cobra.Command, args []string) {
+		config, err := client.GetProjectConfigFromDisk()
+		if err != nil {
+			log.WithError(err)
+		}
+		remotename := strings.Split(cmd.Parent().Use, " ")[0]
```
PiggySpeed: Convert to camelCase to stay consistent with project style?

deployment.go
PiggySpeed: If you do find that second bug that @bobheadxi mentioned, perhaps consider applying the fix through a separate commit to make reviewing easier?

bobheadxi: Heads up that I actually (partially) fixed this in the branch I am working on (`rob/auth` or something) 😅 

Sorry but that's another massive, over-scoped PR coming 😞

bobheadxi: The problem was that `log.WithError()` doesn't stop execution, so the program ends up using things like `resp` from the above example, which of course is `nil` because an `err` was returned.

Issue #137, opened 2018-03-24
Web app login page


bobheadxi:

PiggySpeed:
also: 
- remove dependencies
- simplify & modernize build system

PR #139, opened 2018-03-25
User credentials, authentication, and session control

### LINKED ISSUE #127: Web interface authentication
simply copy/paste the daemon key as password? store as cookie for requests?

bobheadxi:
:tickets: **Ticket(s)**: Closes #127

---

- 📦 Credential management with [boltdb](https://github.com/boltdb/bolt)

- 💻 CLI commands for adding and removing users + resetting database via `userManager`

- 🏳️‍🌈 `bcrypt` password hashing (doesn't seem to need salt? not sure how it works)

- 🚌 Custom session management in `userManager`

- 🔒 Custom permission restriction middleware in `PermissionsHandler`

- ⚒ Minor refactoring throughout `daemon/auth` package

- 📖 README updates, improved development tips

User commands:

```
inertia $VPS users --help
```

Tests:

```
make test
```

I highly recommend not running the tests yet though, there seems to be a major memory leak in `users_test.go` 😛 

## To Do

- [ ] Fix leak in `users_test.go` 🚰 
- [ ] #137 log in via web app ✨


Dockerfile
```diff
@@ -7,7 +7,9 @@ ENV BUILD_HOME=/go/src/github.com/ubclaunchpad/inertia/daemon/web
 ADD ./daemon/web ${BUILD_HOME}
 WORKDIR ${BUILD_HOME}
 # Build and minify client.
-RUN npm install --production
+RUN if [ ! -d "node_modules" ]; then \
```
chadlagore: What if node_modules is out of date?

Makefile
```diff
@@ -9,6 +9,10 @@ RELEASE = canary
 
 all: inertia
 
+# List all commands
+commands:
+	@$(MAKE) -pRrq -f $(lastword $(MAKEFILE_LIST)) : 2>/dev/null | awk -v RS= -F: '/^# File/,/^# Finished Make data base/ {if ($$1 !~ "^[#.]") {print $$1}}' | sort | egrep -v -e '^[^[:alnum:]]' -e '^$@$$' | xargs
```
chadlagore: That’s a good one 😅

README.md
```diff
@@ -64,65 +63,62 @@ $> inertia $VPS_NAME status
...
 - `canary` denotes experimental builds used for testing and development - do not use this.
 
 The daemon component of an Inertia release can be patched separately from the CLI component - see our [wiki](https://github.com/ubclaunchpad/inertia/wiki/Daemon-Releases) for more details.
 
-### Swag
-
-[![Deployed with Inertia](https://img.shields.io/badge/Deploying%20with-Inertia-blue.svg)](https://github.com/ubclaunchpad/inertia)
+## Swag
 
+Add some bling to your Inertia-deployed project :sunglasses:
```
chadlagore: This is so cool

client/deployment.go
```diff
@@ -91,10 +91,31 @@ func (d *Deployment) Logs(stream bool, container string) (*http.Response, error)
...
+	reqContent := &common.UserRequest{Username: username}
+	return d.request("/web/removeuser", "POST", reqContent)
+}
+
+// ResetUsers resets all users on the remote.
+func (d *Deployment) ResetUsers() (*http.Response, error) {
+	return d.request("/web/resetusers", "POST", nil)
+}
+
+func (d *Deployment) request(endpoint, method string, requestBody interface{}) (*http.Response, error) {
```
chadlagore: Idiomatically, aren’t these request arguments normally in the order:

method, endpoint, body...

Dockerfile
bobheadxi: Same problem with `dep` as well, where I skip the dependency install if `vendor` is already there. There's a bit of a risk, but I think the build time improvements is worth it for dev (`npm install` + `dep ensure` together can take more than a few minutes from home 😅 )

Plus, releases are built in a sterile environment on Travis with fresh `node_modules`, so I don't *think* it'll be a problem 🤔

Makefile
bobheadxi: StackOverflow ❤️

client/deployment.go
bobheadxi: Good point, I'll switch it around 👍 

Appreciate the review! 🙏

daemon/inertia/auth/password.go
```diff
@@ -0,0 +1,59 @@
...
+
+func correctPassword(hash string, password string) bool {
+	return bcrypt.CompareHashAndPassword([]byte(hash), []byte(password)) == nil
+}
+
+func validateCredentialValues(username, password string) error {
+	if username == password {
+		return errSameUsernamePassword
+	}
+	if len(password) < 5 {
```
PiggySpeed: also validate username such that it cannot be empty?

daemon/inertia/auth/password.go
```diff
@@ -0,0 +1,59 @@
...
+		return errInvalidPassword
+	}
+	validChars := "abcdefghijklmnopqrstuvwxyzæøåABCDEFGHIJKLMNOPQRSTUVWXYZÆØÅ_0123456789"
+
```
PiggySpeed: what are all these extra characters (æøå)?

daemon/inertia/auth/password.go
PiggySpeed: should we also impose a maximum on password and username to prevent overflows?

daemon/inertia/auth/password.go
bobheadxi: Not sure, they were in a random "list of valid characters" I found - I get rid of the weirder ones

daemon/inertia/auth/password.go
bobheadxi: Good point on making sure the username isn't empty! I'll fix that

As for imposing a maximum on credentials, I feel like if there really is an overflow, it would already be a problem before we reach this point, when we try to parse the request body - that said, perhaps there should be a maximum regardless. 100 characters? 🤔

PiggySpeed: Yeah in all likelihood it would crash something else, but I think it's good design to not rely on that. Possibly set it to 128 as per [owasp]?(https://www.owasp.org/index.php/Authentication_Cheat_Sheet#Password_Length)

Also related discussion on passwords: [stackoverflow](https://stackoverflow.com/questions/98768/should-i-impose-a-maximum-length-on-passwords)

bobheadxi: I suppose it might depend on the hashing algorithm - I'll look into it, but 128 seems like a reasonable limit. Thanks for pointing this out!

daemon/inertia/auth/password.go
```diff
@@ -0,0 +1,64 @@
...
+	if len(password) < 5 || len(password) >= 128 {
+		return errInvalidPassword
+	}
+	if len(username) < 3 || len(username) >= 128 {
+		return errInvalidUsername
+	}
+	validChars := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_"
+
+NEXT_USERNAME_CHAR:
+	for _, char := range username {
```
iKevinY: Not sure how idiomatic this is, but I would avoid the nested loops here using a helper function:

```golang
// Returns true if `str` only contains characters [A-Z], [a-z], or '_'
func legalString(str string) bool {
	for _, c := range str {
		if (c < 'a' || c > 'z') && (c < 'A' || c > 'Z') && c != '_' {
			return false
		}
	}
	return true
}
```

Then the validity checks can look like this:

```golang
if len(password) < 5 || len(password) >= 128 || !validString(password) {
	return errInvalidPassword
}
if len(username) < 3 || len(username) >= 128 || !validString(username) {
	return errInvalidUsername
}
```

daemon/inertia/auth/password_test.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func TestHashPassword(t *testing.T) {
+	unhashed := "amazing"
+	hashed, err := hashPassword(unhashed)
+	assert.Nil(t, err)
+	assert.NotEqual(t, unhashed, hashed)
+}
+
+func TestCorrectPassowrd(t *testing.T) {
```
iKevinY: Minor typo: `s/Passowrd/Password/g`

daemon/inertia/auth/util.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"io"
+)
+
+func generateSessionID() string {
+	b := make([]byte, 32)
+	if _, err := io.ReadFull(rand.Reader, b); err != nil {
+		return ""
```
iKevinY: Seems like the fallback in the (extremely rare) case that we fail to read from `rand.Reader` should be to throw some sort of error, rather than defaulting to setting the session ID to the empty string.

daemon/inertia/auth/password.go
bobheadxi: Wow this is a very clean alternative, thanks!

daemon/inertia/auth/util.go
bobheadxi: I was a bit hesitant to add a `error` return on such a trivial function that should practically never fail, since `rand.Reader` should be safe to rely on and `b` is allocated within the same function and not passed in from elsewhere, but I suppose it's just a few extra lines of `err` checking

Dockerfile
brian-nguyen: You can try adding a step that only copies over the package.json only, then running npm install, and then copying over your source files. If I'm understanding Docker correctly ( 😅), your node_modules + package.json will be cached in a layer and npm install will only run if package.json has changed rather than anything in your source code.

bobheadxi: Nice! A bit better than the current approach in that docker will automatically keep the modules up to date it needed. I imagine something similar with dep can be done as well

bobheadxi:
Thanks @PiggySpeed, but it might be good to leave this open until we resolve the login page issues (#151)

bobheadxi:
@PiggySpeed finally FINALLY got everything in regards to web app authentication working right oh my lawd I'm gonna merge this and start working on logs ASAP but take a look when you can

PR #126, opened 2018-03-18
Inertia Web scaffold

bobheadxi:
:tickets: **Ticket(s)**: Closes #74

---

## :construction_worker: Changes

* Added daemon/web component as a React app
* Updated Dockerfile to build web component
* Updated daemon to serve a super simple web app at /admin/

screen shot 2018-03-18 at 3 08 36 pm

## To Do
- [x] (fixed!) serves `status 404 not found` 😢 both on https://0.0.0.0:8081/admin/inertia-web/ and https://0.0.0.0:8081/admin/ 🤔  
- [x] Pointing PR to rob/#103-packaging for the time being, change to master after that gets merged

## :flashlight: Testing Instructions

Locally:
```
make web-dev
```

Or:
```bash
make testdaemon
# in project...
inertia local init
# visit https://0.0.0.0:8081/admin/ and ignore the cert warning
```

PR #151, opened 2018-04-03
Web interface login

### LINKED ISSUE #137: Web app login page

PiggySpeed:
:tickets: **Ticket(s)**: Closes #137

---

A brief summary of what changes were introduced.

- simplified web app dependencies and webpack config
- attempt at connecting with daemon (will need some help from @bobheadxi for this)

### DIFF
--- a/daemon/web/package-lock.json
+++ b/daemon/web/package-lock.json
@@ -7,14 +7,12 @@
     "@sindresorhus/is": {
       "version": "0.7.0",
       "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.7.0.tgz",
-      "integrity": "sha512-ONhaKPIufzzrlNbqtWFFd+jlnemX6lJAgq9ZeiZtS7I1PIf/la7CW4m83rTXRnVnsMbW2k56pGYu7AUFJD9Pow==",
-      "dev": true
+      "integrity": "sha512-ONhaKPIufzzrlNbqtWFFd+jlnemX6lJAgq9ZeiZtS7I1PIf/la7CW4m83rTXRnVnsMbW2k56pGYu7AUFJD9Pow=="
     },
     "accepts": {
       "version": "1.3.5",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
       "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
-      "dev": true,
       "requires": {
         "mime-types": "2.1.18",
         "negotiator": "0.6.1"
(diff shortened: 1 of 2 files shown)


daemon/web/components/App.js
```diff
@@ -6,29 +6,74 @@ import InertiaClient from '../client';
...
+      password: "",
+      loginAlert: ""
+    };
+    this.handleLoginSubmit = this.handleLoginSubmit.bind(this);
+    this.handleUsernameBlur = this.handleUsernameBlur.bind(this);
+    this.handlePasswordBlur = this.handlePasswordBlur.bind(this);
+  }
+
+  async handleLoginSubmit() {
+    const endpoint = 'https://localhost:8081/web/login';
```
bobheadxi: Note that `client.js` has the [daemon host path](https://github.com/ubclaunchpad/inertia/blob/john/%23137-web-login/daemon/web/client.js#L5) which is [dynamically determined](https://github.com/ubclaunchpad/inertia/blob/john/%23137-web-login/daemon/web/index.js#L8). The client is then [passed to the App](https://github.com/ubclaunchpad/inertia/blob/john/%23137-web-login/daemon/web/index.js#L10).

```js
const daemonAddress = window.location.host;
const client = new InertiaClient(daemonAddress);
ReactDOM.render(<App client={client}/>, document.getElementById('app'));
```

Instead of having this function here, i think we should place all API calls in `client.js` and access it through the `InertiaClient` class - this allows us to avoid having the define endpoints everywhere

Another note: I don't think `localhost` works very well because of the way the Docker routing works, have you tried using `0.0.0.0:8081/web/login`? See #126

bobheadxi:
Another thought, the web app shouldn't even be accessible at the moment because of [how I've set up the permissions handler](https://github.com/ubclaunchpad/inertia/blob/rob/%23127-authentication/daemon/inertia/daemon.go#L78):

```go
	// Inertia web - PermissionsHandler is used to authenticate web
	// app access and manage users
	webPrefix := "/web"
	permHandler, _ := auth.NewPermissionsHandler(
		auth.UserDatabasePath, host, webPrefix, 120,
	)
	defer permHandler.Close()
	permHandler.AttachUserRestrictedHandler(
		"/", http.FileServer(http.Dir("/app/inertia-web")),
	)

        // everything served from '/web' should be restricted to already logged in users
	mux.Handle(webPrefix, http.StripPrefix(webPrefix, permHandler))
```

do I write bad code? 😅

daemon/web/components/App.js
PiggySpeed: I modified the code to use the `daemonAddress`.

However, `window.location.host` becomes `http://localhost:7900` when I run it in dev mode. So I've added a `NODE_ENV` environment variable to webpack as a temporary measure for now.

PiggySpeed: I suppose an option would be to have `make web-build` pass along the daemon url as an env variable.

bobheadxi:
PermissionsHandler has a function to add exceptions that don’t need a logged in user - these routes won’t require an active session

daemon/web/components/App.js
bobheadxi: I think the way you set it up as is fine (with the node env)

arjansandhu:
lookin good so far!

daemon/web/common/AuthService.js
```diff
@@ -0,0 +1,38 @@
...
+  return true;
+};
+
+export const guardRoute = () => {
+  if (!isAuthenticated()) {
+    // push route
+  }
+};
+
+export const setToken = (token) => {
```
bobheadxi: The daemon should be setting cookies itself - I don't think you have to manually manage tokens.

[Setting cookie upon login](https://github.com/ubclaunchpad/inertia/blob/rob/%23127-authentication/daemon/inertia/auth/sessions.go#L90)

[Removing cookies](https://github.com/ubclaunchpad/inertia/blob/rob/%23127-authentication/daemon/inertia/auth/sessions.go#L122)

[Tests verify that cookies are received (I hope)](https://github.com/ubclaunchpad/inertia/blob/rob/%23127-authentication/daemon/inertia/auth/permissions_test.go#L145)

PiggySpeed: Ah i see, I was wondering how we are handling this.

PR #154, opened 2018-04-05
Fix Docker bootstrapping script

### LINKED ISSUE #153: Docker bootstrapping fails on `inertia [VPS] init`
I tried running `inertia $VPS init` and got the following error:

```bash
$ inertia <vps> init
Setting up remote <vps> at <remote>
>> Step 1/4: Installing docker...
bash: line 30: !(hash docker 2>/dev/null): No such file or directory
Failed to start docker.service: Unit docker.service not found.

FATA[0003] Process exited with status 5
```

The problem is that [`!(hash docker 2>/dev/null)`](https://github.com/ubclaunchpad/inertia/blob/72733b74a252a054478e9212caf8f6e193264e6b/client/bootstrap/docker.sh#L31) is actually doing some strange pseudo-`ls` 🤔:

```
$ ls
$ echo !(hash docker 2>/dev/null)
!(hash docker 2>/dev/null)
$ touch a.txt
$ echo !(hash docker 2>/dev/null)
a.txt
$ touch b.txt
$ echo !(hash docker 2>/dev/null)
a.txt b.txt
```

Since we do an early exit when Docker is already installed here, I think the problematic conditional can just be removed altogether?

https://github.com/ubclaunchpad/inertia/blob/72733b74a252a054478e9212caf8f6e193264e6b/client/bootstrap/docker.sh#L13-L15

iKevinY:
:tickets: **Ticket(s)**: Closes #153

---

Fixes a problem with Docker installation during `inertia init` bootstrapping.

Figure out what circumstances caused the bug in the first case, and then try to reproduce it. 😇


chadlagore:
:tada: The test suite may actually run this chunk of code on Travis, it does some cool integration tests. @bobheadxi would know.

chadlagore:
Ah, you need to run `go get -u github.com/jteeuwen/go-bindata/...` and `make bootstrap` and push the resulting go files to make the tests pass. That should happen automatically I think. Maybe another ticket is in order 🤔

bobheadxi:
Yep, there is one integration test that runs through the bootstrap on a number of platforms!

https://github.com/ubclaunchpad/inertia/blob/master/client/remote_test.go#L113

RE: having `make bootstrap` run automatically, I'm not super sure it's feasible, since it requires and additional commit, and I'm not too sure how I feel about commit automation 🤔 Just something we have to remember to do.

@iKevinY instructions [here](https://github.com/ubclaunchpad/inertia#compiling-bash-scripts)

iKevinY:
@chadlagore Does this look good to you, or should we change the Docker check altogether?

bobheadxi:
I think the Docker check as is, is actually better - pointed out in #153 that we won't be able to start the Docker daemon without the binary

PR #191, opened 2018-05-05
Modify configuration from CLI

### LINKED ISSUE #111: Modify remotes in CLI
```
inertia remote set $VPS $PROPERTY $VALUE
```

rogermyang:
:tickets: **Ticket(s)**: Closes #111 

---

Added set remote CLI functionalities. Currently user has to enter the exactly field name (i.e. Name,IP,User,PEM,Branch) to change the string fields of the remotes. 

Inertia configuration fields can be modified as well. This works for any string field such as Version,Project and BuildType.

inertia remote set [REMOTE] [FIELD] [VALUE]


client/config.go
```diff
@@ -71,6 +73,18 @@ func (config *Config) RemoveRemote(name string) bool {
 	}
 	return false
 }
+func (config *Config) SetProperty(name string, value string) bool {
+	f := reflect.ValueOf(config).Elem().FieldByName(name)
+	if f.IsValid() {
```
bobheadxi: From the `reflect` documentation:

> IsValid reports whether v represents a value. It returns false if v is the zero Value.

Does this mean that if the config currently has no value set for a field, you won't be able to set it? I think if the field is `null` we should still allow users to set a value to it.

client/config.go
```diff
@@ -175,3 +189,9 @@ func GetConfigFilePath() (string, error) {
 	}
 	return filepath.Join(path, configFileName), nil
 }
+
+func (config *Config) PrintConfigDetails() {
```
bobheadxi: This probably shouldnt be here - the CLI should handle this, similarly to how remote details are currently printed: https://github.com/ubclaunchpad/inertia/blob/master/remote.go#L121

client/remote.go
```diff
@@ -188,3 +189,14 @@ func (remote *RemoteVPS) getDaemonAPIToken(session SSHSession, daemonVersion str
 	// There may be a newline, remove it.
 	return strings.TrimSuffix(stdout.String(), "\n"), nil
 }
+
+func (remote *RemoteVPS) SetRemoteProperty(name string, value string) bool {
```
bobheadxi: Not sure if this function is needed - see other comment on `config.SetProperty()`

remote.go
```diff
@@ -137,6 +137,42 @@ var showCmd = &cobra.Command{
...
+
+		remote, found := config.GetRemote(args[0])
+		if found {
+			success := config.SetProperty(args[1], args[2])
+			if success {
+				println("Configuration has been updated.")
+				config.PrintConfigDetails()
+				printRemoteDetails(remote)
+
+			} else {
```
bobheadxi: What happens if I want to change the Inertia version in the config file?
```
inertia set my_vps version v0.2.0
```
But you're not actually updating something about `my_vps` - the argument doesn't actually mean anything here.

Perhaps a better command tree would be:
```
inertia set [config property] [config value]
```
and:
```
inertia [REMOTE] set [remote property] [remote value]
```

client/config.go
rogermyang: no IsValid() is checking whether a field(key) is found is the struct. The implementation of that is checking a flag resides in the value struct.

client/config.go
rogermyang: i see ill just remove it

client/config.go
bobheadxi: ohh I see

remote.go
bobheadxi: This way you also won't have to do this double loop (check for presence of property in `config`, then check for it in `remote`)

client/remote.go
```diff
@@ -188,3 +189,14 @@ func (remote *RemoteVPS) getDaemonAPIToken(session SSHSession, daemonVersion str
...
+	f := reflect.ValueOf(remote).Elem().FieldByName(name)
+	if f.IsValid() {
```
bobheadxi: This code:
```go
	f := reflect.ValueOf(remote).Elem().FieldByName(name)
	if f.IsValid() {
		if f.Kind() == reflect.String {
			f.SetString(value)
			return true
		}
	}
	return false
```

Is repeated twice. It could probably be abstracted into a helper function that both `SetProperty` and `SetRemoteProperty` can rely on:

```go
func SetStructProperty(name string, value string, struct interface{}) { /* ... */ }
```

remote.go
rogermyang: good point I was thinking of making two commands as well.

client/config.go
```diff
@@ -71,6 +72,31 @@ func (config *Config) RemoveRemote(name string) bool {
 	}
 	return false
 }
+func SetProperty(name string, value string,structObject interface{}) bool {
```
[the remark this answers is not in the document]
rogermyang: Thanks Rob for review and feedback. Yes I should probably just switch to VS code. There's some clean up work and test I was going to add so this push is still WIP. Right now I think I'll fix and issues and cleaning up before pushing more.

bobheadxi:
oh my god Travis just pass the damned build please 😡

PR #192, opened 2018-05-10
#176 querystring (for GET requests)

### LINKED ISSUE #176: Standardize GET requests
Well TIL that we shouldn't use Bodies with `GET` requests. Go's `client.Do()` doesn't complain, but Inertia Web complains if you `GET` with a `Body` 😢 

Looks like GET requests should have their options in the URL. This might actually work out in our favour, simplifying the `DaemonRequest` struct

Daemon endpoints will need to be update to use `url.Query()`

PiggySpeed:
:tickets: **Ticket(s)**: Closes #176 
---

- requests are now split into separate get/post functions in the `deployment.go` file
- GET requests now process query strings
- daemon endpoints updated accordingly
- changes are only relevant to the logging functionality

- run tests as usual

### DIFF
--- a/client/deployment.go
+++ b/client/deployment.go
@@ -10,11 +10,11 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strconv"
 	"strings"
 
 	"github.com/ubclaunchpad/inertia/common"
 	git "gopkg.in/src-d/go-git.v4"
-	"strconv"
 )
 
 // Deployment manages a deployment
@@ -103,13 +103,9 @@ func (d *Deployment) Reset() (*http.Response, error) {
 
 // Logs get logs of given container
 func (d *Deployment) Logs(stream bool, container string) (*http.Response, error) {
-	//reqContent := &common.DaemonRequest{
-	//	Stream:    stream,
-	//	Container: container,
-	//}
 	reqContent := map[string]string{
-		"Stream": strconv.FormatBool(stream),
-		"Container": container,
+		common.Stream:    strconv.FormatBool(stream),
+		common.Container: container,
 	}
 
 	return d.get("/logs", reqContent)
@@ -143,16 +139,8 @@ func (d *Deployment) ListUsers() (*http.Response, error) {
 
 // Sends a GET request. "queries" contains query string arguments.
 func (d *Deployment) get(endpoint string, queries map[string]string) (*http.Response, error) {
-	// Assemble URL
-	url, err := url.Parse("https://" + d.RemoteVPS.GetIPAndPort())
-	if err != nil {
-		return nil, err
-	}
-	url.Path = path.Join(url.Path, endpoint)
-	urlString := url.String()
-
 	// Assemble request
-	req, err := http.NewRequest("GET", urlString, nil)
+	req, err := d.buildRequest("GET", endpoint, nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


PiggySpeed:
I'll look into why build is failing...

bobheadxi:
Looks like it's just the linter, should be a quick fix :)
```
client/deployment.go:1::warning: file is not gofmted with -s (gofmt)
client/deployment.go:1::warning: file is not goimported (goimports)
daemon/inertia/logs.go:1::warning: file is not goimported (goimports)
```

https://travis-ci.org/ubclaunchpad/inertia/jobs/377142082#L1175

client/deployment.go
```diff
@@ -97,16 +98,21 @@ func (d *Deployment) Status() (*http.Response, error) {
...
-	return d.request("POST", "/reset", nil)
+	return d.post("/reset", nil)
 }
 
 // Logs get logs of given container
 func (d *Deployment) Logs(stream bool, container string) (*http.Response, error) {
-	reqContent := &common.DaemonRequest{
-		Stream:    stream,
-		Container: container,
+	//reqContent := &common.DaemonRequest{
```
bobheadxi: Let's leave out the commented code 🚽

client/deployment.go
```diff
@@ -97,16 +98,21 @@ func (d *Deployment) Status() (*http.Response, error) {
...
+	//	Stream:    stream,
+	//	Container: container,
+	//}
+	reqContent := map[string]string{
+		"Stream": strconv.FormatBool(stream),
```
bobheadxi: isn't it standard for request parameters to be lowercased?

client/deployment.go
```diff
@@ -116,26 +122,67 @@ func (d *Deployment) AddUser(username, password string, admin bool) (*http.Respo
...
+	// Make HTTPS request
+	tr := &http.Transport{
+		// Our certificates are self-signed, so will raise
+		// a warning - currently, we ask our client to ignore
+		// this warning.
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+		},
+	}
+	client := &http.Client{Transport: tr}
```
bobheadxi: The HTTP clients (and transports) are the same for all request types - could we have it set up and fetched from a helper function?

client/deployment.go
```diff
@@ -116,26 +122,67 @@ func (d *Deployment) AddUser(username, password string, admin bool) (*http.Respo
...
+	req, err := http.NewRequest("GET", urlString, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add query strings
+	if queries != nil {
+		q := req.URL.Query()
+		for k, v := range queries {
+			q.Add(k, v)
```
bobheadxi: Nice!

client/deployment_test.go
```diff
@@ -205,14 +205,8 @@ func TestLogs(t *testing.T) {
...
 		// Check body
-		body, err := ioutil.ReadAll(req.Body)
-		assert.Nil(t, err)
 		defer req.Body.Close()
-		var upReq common.DaemonRequest
-		err = json.Unmarshal(body, &upReq)
-		assert.Nil(t, err)
-		assert.Equal(t, "docker-compose", upReq.Container)
-		assert.Equal(t, true, upReq.Stream)
+		assert.Equal(t, "Container=docker-compose&Stream=true", req.URL.RawQuery)
```
bobheadxi: Since the daemonside handler uses this syntax:
```go
container := q.Get("Container")
```
could we use that here in the test as well, just to ensure consistency?

I also wonder if the argument names (`container` , `stream`) might be better to have as exported variables in package `common` alongside the Request structs?

bobheadxi:
Also note that you'll want to update [some of the Inertia Web API requests](https://github.com/ubclaunchpad/inertia/blob/master/daemon/web/client.js#L37) as well 💻

client/deployment.go
PiggySpeed: done

bobheadxi: I know commenting "done" is pretty popular, but tbh just pushing a commit that fixes this is usually enough, GitHub will mark this code as changed 😆 idm either way though

PiggySpeed: oh sorry, I usually do it to keep track and make sure I've addressed all comments! Are you getting emails every time I do that?

bobheadxi: I don't have email notifications enabled no, so it's all g either way 👍

client/deployment_test.go
PiggySpeed: Good point! Yeah I agree that it would make sense to put them in a common folder, since they're used across different files.

PiggySpeed:
Alright, I think it's all good for review @bobheadxi

PR #196, opened 2018-05-12
Faster deploys

### LINKED ISSUE #173: Faster repository deploy when updating
All 3 builds are two-staged (build then deploy) - hold off on container shut down until build completes when deploying

bobheadxi:
:tickets: **Ticket(s)**: Closes #173

---

This PR refactors some of the build code, adds some tests, and changes the deploy process so that a build is allowed to complete before the old containers are shut down and the new containers are brought up.

Some new interfaces:

```go
// Builder builds projects and returns a callback that can be used to deploy the project.
type Builder func(*Deployment, *docker.Client, io.Writer) (func() error, error)

type containerStopper func(*docker.Client, io.Writer) error
```

```
make test-all
```

Also needs some further testing with real projects, since I'm not sure if container conflicts will come into play. There are tests, but I'm not sure how well they reflect real usage since no actual code changes occur.

### DIFF
--- a/daemon/inertia/project/build.go
+++ b/daemon/inertia/project/build.go
@@ -28,6 +28,10 @@ const (
 	BuildStageName = "build"
 )
 
+// Builder builds projects and returns a callback that can be used to deploy the project.
+// No relation to Bob the Builder, though a Bob did write this.
+type Builder func(*Deployment, *docker.Client, io.Writer) (func() error, error)
+
 // getTrueDirectory converts given filepath to host-based filepath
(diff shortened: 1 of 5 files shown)


daemon/inertia/project/build.go
```diff
@@ -28,6 +28,10 @@ const (
 	BuildStageName = "build"
 )
 
+// Builder builds projects and returns a callback that can be used to deploy the project.
+// No relation to Bob the Builder, though a Bob did write this.
```
PiggySpeed: 🤣

PR #198, opened 2018-05-13
John/#188 inertia web linting

### LINKED ISSUE #188: Set Up Linting for Inertia Web
This task involves three steps:

1. Select and install a linting solution for React/JS code. I suggest [AirBnB](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb)'s guidelines as it is currently the most popular solution and enforces more appropriate (imo) rules than the alternatives. A brief review of alternatives can be found [here](https://medium.com/@uistephen/style-guides-for-linting-ecmascript-2015-eslint-common-google-airbnb-6c25fd3dff0).

2. Fix all existing codebase linting violations.

3. Automate the linting process via webpack.

PiggySpeed:
:tickets: **Ticket(s)**: Closes #188 

---

Implemented ESLint, fixed all linting errors.

We will adjust linting rules as the web project matures.

You can now run `npm lint` to lint intertia web.


bobheadxi:
Always seems to be the docker-compose test failing:

```
Starting up project...
--- FAIL: TestDockerComposeIntegration (30.28s)
	Error Trace:	build_test.go:115
	Error:		Should be true
	Messages:	project container should be active
```

looks like the timeout still isn't generous enough, i'll increase it in one of the other PRs 😶

daemon/web/package.json
```diff
@@ -9,27 +9,34 @@
   "scripts": {
     "start": "NODE_ENV=development webpack-dev-server --host=127.0.0.1",
     "build": "NODE_ENV=production webpack --config webpack.prod.js",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "lint": "eslint . --ignore-pattern public/"
   },
   "dependencies": {
-    "babel-core": "^6.26.0",
+    "babel-core": "^6.26.3",
```
bobheadxi: Let's pin down production dependencies to specific versions:

```js
"babel-core": "6.26.3"
```

daemon/web/.eslintrc.json
```diff
@@ -0,0 +1,22 @@
+{
```
bobheadxi: Small thing, but I feel like there is some advantage to keeping this in the root directory alongside other general configuration (as it used to be) - unless this has to be in the same directory as package.json and stuff

PiggySpeed: It would make sense to have all configuration/linting files in one place, but I'm still feeling a bit reluctant to do that. I generally promote files up a level it if needs to know of files from that directory. I'm also inclined to enforce a stronger separation between the web-app project and the daemon/cli. Also, [this](https://github.com/Microsoft/vscode-eslint/issues/196).

Makefile
```diff
@@ -29,11 +29,12 @@ inertia-tagged:
 
 # Remove Inertia binaries
 clean:
-	rm -f ./inertia 
+	rm -f ./inertia
 	find . -type f -name inertia.\* -exec rm {} \;
 
 lint:
 	PATH=$(PATH):./bin bash -c './bin/gometalinter --vendor --deadline=60s ./...'
+	cd ./daemon/web; npm run lint
```
bobheadxi: Hard-cd-ing into `./daemon/web` might have weird consequences if more commands are added, etc.

Might want to use something similar to the [other web commands](https://github.com/ubclaunchpad/inertia/blob/b92709169e4f29d091a29653f4b1d229d1756d39/Makefile#L105):

```bash
web-run:
   (cd ./daemon/web; npm start)
```

This will cd into the directory only for that particular command

PiggySpeed: Haha, no worries, feel free to add in as many comments as you want.

That makes a lot more sense! Will make the change.

Issue #204, opened 2018-05-19
Inertia Web 1.0


bobheadxi:
This project will involve a complete redesign and rewrite of Inertia Web to bring it to feature parity with the command line application and enable improved team features for the Inertia platform. Inertia branding will probably also be updated.

## Roadmap

- [x] Wireframe web app and update Inertia branding (#160)
- [x] Scaffold implementation of updated design (#248)
- [ ] Add functionality until feature parity with CLI is achieved

bobheadxi:
Closing as this is now a dupe of #248, opened by @PiggySpeed

bobheadxi:
Reopening because this is not quite a dupe :p

bobheadxi:
closing, as we are moving away from zenhub

PR #220, opened 2018-05-30
WebSockets

### LINKED ISSUE #174: Websockets instead of plain requests
So we can get things like live build status, etc as well as a cleaner log implementation

However, not sure if actually needed

bobheadxi:
:tickets: **Ticket(s)**: Closes #174 

---

This PR significantly refactors the daemon's logging stuff and switches `inertia [remote] logs --stream` to use websockets instead of the old bit streaming.

I don't think the other commands (`inertia [remote] up --stream`) will be migrated to this since they actually work a bit better with the old streamer, and you can't POST to initiate a websocket connection.  I think `inertia [remote] ssh` will be nice to have on websockets, but I'll leave that for another PR.

I don't think I will update Inertia web for this change for now, since that is pending a rewrite soon (#204)

This is more or less ready for review, though I need to do some more thorough testing and make sure everything is tidy.

As always:

```
make test
```

You can also set up a deployment and try out the logs via `inertia [remote] logs`.


bobheadxi:
What is wrong with you coveralls :(

PR #224, opened 2018-05-31
Brian/#223 dev proxy

### LINKED ISSUE #223: Proxy requests in Dev to avoid CORS
Since Webpack is making requests from a different domain (localhost:7900) to the daemon, CORS was enabled.

Webpack Dev Server can proxy its requests through the daemon to avoid this issue.

brian-nguyen:
:tickets: **Ticket(s)**: Closes #223 

---

CORS only affects development due to Webpack Dev Server. Set up proxy config in package.json and webpack [config](https://webpack.js.org/configuration/dev-server/#devserver-proxy), and then removed CORS handling. As a result, all requests to the daemon are now made relative to the server root.

Open the Webpack hosted page. Note that for testing, I disabled the cookie and admin checks on the daemon to isolate whether the `/validate` route was correctly proxied or not. Some more help is needed!

### DIFF
--- a/daemon/inertiad/auth/permissions.go
+++ b/daemon/inertiad/auth/permissions.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"net/url"
 	"strings"
 
 	jwt "github.com/dgrijalva/jwt-go"
@@ -88,19 +87,6 @@ func (h *PermissionsHandler) Close() error {
 
 // nolint: gocyclo
 func (h *PermissionsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	// CORS for development
-	origin, err := url.Parse(r.Header.Get("Origin"))
-	if err == nil && origin.Hostname() == "127.0.0.1" {
-		allowedOrigin := "http://" + origin.Host
-		w.Header().Set("Access-Control-Allow-Origin", allowedOrigin)
-		w.Header().Set("Access-Control-Allow-Credentials", "true")
-		w.Header().Set("Access-Control-Allow-Methods", "POST OPTIONS")
-		w.Header().Set("Access-Control-Allow-Headers", "Content-Type")
-		if r.Method == "OPTIONS" {
-			return
-		}
-	}
-
 	// http.StripPrefix removes the leading slash, but in the interest
 	// of maintaining similar behaviour to stdlib handler functions,
 	// we manually add a leading "/" here instead of having users not add
--- a/daemon/web/client.js
+++ b/daemon/web/client.js
@@ -5,17 +5,17 @@ export default class InertiaClient {
     this.url = 'https://' + url;
   }
 
-  async logout() {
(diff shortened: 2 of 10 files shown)


brian-nguyen:
Removal of `this.url` from the get and post methods in [client.js](https://github.com/ubclaunchpad/inertia/blob/c78680f4c6a21842ceeb77e3af1e454357cebd12/daemon/web/client.js#L76) has resulted in eslint complaining. How should this be handled? Is the rule `class-methods-use-this` something that is strictly enforced?

daemon/inertiad/auth/permissions.go
```diff
@@ -88,19 +87,6 @@ func (h *PermissionsHandler) Close() error {
 
 // nolint: gocyclo
 func (h *PermissionsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	// CORS for development
-	origin, err := url.Parse(r.Header.Get("Origin"))
```
bobheadxi: nice nice this was ugly 😅

daemon/web/webpack.config.js
```diff
@@ -13,6 +13,9 @@ const config = {
     inline: true,
     contentBase: './public',
     publicPath: '/web/',
+    proxy: {
+      '/': { target: 'https://127.0.0.1:4303', secure: false },
```
bobheadxi: TIL 🔥

PiggySpeed:
@brian-nguyen  @bobheadxi 

> Removal of this.url from the get and post methods in client.js has resulted in eslint complaining. How should this be handled? Is the rule class-methods-use-this something that is strictly enforced?

This warning is suggesting that we may want to convert the `get` method into a static because it does not depend on any information from the instance. In fact, by removing `this.url` from the `get` and `post` methods, we are not making use of any instance variables and can convert the whole thing to a bunch of module-level helper functions.

So here are some options:
1. remove the constructor, and convert all the methods to static ones; remove all uses of `this` in all other methods
2. extract all methods out into module-level helper functions (i.e. remove the class altogether)
I am okay with either one.

When it comes to eslint rules, I play on the conservative side and try to avoid changing them as much as possible.

PiggySpeed:
> I think the web app is due for an overhaul anyway

Yes, we don't have to be very picky about the elegance of the codebase since the existing web app was just a quick prototype until we fleshed out more details about the interface.

daemon/web/client.js
```diff
@@ -66,14 +66,15 @@ export default class InertiaClient {
    * @param {Object} params
    * @param {{[key]: string}} queryParams
    */
+  // eslint-disable-next-line
```
PiggySpeed: See previous comment regarding `this.url`.

brian-nguyen:
@PiggySpeed @bobheadxi 

Sounds good. I just went ahead and made the methods all static and changed all usage to those static methods. Let me know if I missed anything

bobheadxi:
paging @zhulia96 and @rogermyang who will be working on the web app a bit more on this dicussion

Issue #240, opened 2018-06-14
Inertia Web - Global Styles SASS


PiggySpeed:

PiggySpeed:
@asitu you can take a look at this for reference: https://github.com/athappa/cpsc_304/tree/master/client/src/styles

bobheadxi:
@PiggySpeed for me at least that link is inaccessible (404 not found)

asitu:
The link is inaccessible for me too

PiggySpeed:
ah sorry, try this one instead: https://github.com/PiggySpeed/shopify-challenge/tree/master/src/styles

PR #251, opened 2018-06-15
Imported assets + SASS global styles (#240)

asitu:
- Uploaded footer icons, top navigation icons, logos, and font files to an Assets folder and Font folder.
- Added SASS codes to typography.sass and colors.sass


daemon/web/styles/typography.sass
```diff
@@ -0,0 +1,50 @@
+/* Text styles */
+
+=Paragraph()
```
bobheadxi: didn't know this was valid sass syntax 🤔

daemon/web/styles/typography.sass
```diff
@@ -0,0 +1,50 @@
...
+  font-size: 14px
+  font-weight: normal
+  font-style: normal
+  font-stretch: normal
+  line-height: normal
+  letter-spacing: normal
+  color: $greyblue
+
+
+h2
```
bobheadxi: maybe nitpicking but it might be nice if these were in order (`h1` > `h2` > `h3` > ...)

bobheadxi: on that topic, most of the traits on these styles are the same - I think you can fiddle around with inheritance a bit to bring down the verbosity. @PiggySpeed can probably correct me on this, but it might go:

```sass
%header-base
   font-family: Rubik
   font-style: normal
   font-stretch: normal
   line-height: normal

h3
   @extend header-base
   font-size: 14px
   font-weight: 500
   letter-spacing: normal
   color: $darkgrey

...
```

daemon/web/styles/typography.sass
PiggySpeed: @asitu could you provide a link, or explain what this syntax does? I've been trying to Google it, but there's nothing out there.

daemon/web/styles/typography.sass
PiggySpeed: I would also suggest mixins as there are quite a few shared properties.

PiggySpeed:
@asitu I would suggest using Google CDN for loading our fonts, `<link href='...' />`. We should only load the font weights we need. Advantage of CDN is being able to load font file formats that are optimized for the browser.

https://fonts.google.com/specimen/Rubik?selection.family=Rubik

daemon/web/styles/index.sass
```diff
@@ -1,5 +1,6 @@
 @import ./colors
 @import ./typography
+<link href="https://fonts.googleapis.com/css?family=Rubik" rel="stylesheet">
```
bobheadxi: I believe this should be here 😋 

https://github.com/ubclaunchpad/inertia/blob/master/daemon/web/index.html

There are a few others there already, we might be able to remove them for the time being:

```html
<head>
  <meta charset="UTF-8">
  <link href='https://fonts.googleapis.com/css?family=Roboto:400,300,500' rel='stylesheet' type='text/css'>
  <link href="https://fonts.googleapis.com/css?family=Ubuntu+Mono" rel="stylesheet">
  <title>Inertia Web</title>
</head>
```

daemon/web/fonts/Rubik/OFL.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2015 The Rubik Project Authors,
```
bobheadxi: Since you'll be importing the font from online, you should delete the `/Rubik` directory

daemon/web/styles/index.sass
bobheadxi: @asitu make sure you remove this line! It is currently causing builds to fail:

```
@import ./typography
                      ^
      Invalid CSS after '..."./typography";': expected 1 selector or at-rule, was '<link href="https:/'
      in /go/src/github.com/ubclaunchpad/inertia/daemon/web/styles/index.sass (line 2, column 24)
 @ ./styles/index.sass 2:14-117
```

bobheadxi: Once builds are going again I think this is good to merge for the time being - we can work on the cleaning up the stylesheets this weekend

PR #267, opened 2018-06-23
Fix config send, add --dest flag

### LINKED ISSUE #266: Add --dest flag for inertia [remote] send

bobheadxi:
:tickets: **Ticket(s)**: Closes #266

---

- Fix incorrect send destination
- Add `--dest` flag to specify where you want your file to go relative to project root

n/a

### DIFF
--- a/deploy.go
+++ b/deploy.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/ubclaunchpad/inertia/common"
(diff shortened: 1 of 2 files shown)


deploy.go
```diff
@@ -372,8 +373,15 @@ deployment. Provide a relative path to your file.`,
...
-		// Destination path
-		remotePath := path.Join(common.DaemonProjectPath, args[0])
+		// Get flag for destination
+		dest, err := cmd.Flags().GetString("dest")
+		if err != nil {
+			log.Fatal(err.Error())
+		}
+
+		// Destination path - todo: allow config
+		projectPath := "/app/host/inertia/project"
```
brian-nguyen: Quick question: Why does the path start with `/app/host`?

deploy.go
```diff
@@ -84,6 +84,7 @@ Run 'inertia [REMOTE] init' to gather this information.`,
 		cmd.AddCommand(ssh)
 
 		send := deepCopy(cmdDeploymentSendFile)
+		send.Flags().StringP("dest", "d", "", "Path relative from project root to send file to")
```
brian-nguyen: 🔥

deploy.go
bobheadxi: in `client/scripts/daemon-up.sh`, we mount the host home directory to there:

```bash
sudo docker run -d \
    --restart unless-stopped \
    -p "$DAEMON_PORT":"$CONTAINER_PORT" \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v "$HOME":/app/host \ # <--- relevant line
    -e HOME="$HOME" \
    -e SSH_KNOWN_HOSTS='/app/host/.ssh/known_hosts' \
    --name "$DAEMON_NAME" \
    "$IMAGE" "$HOST_ADDRESS" > /dev/null 2>&1
```

so on the host, `$HOME/inertia/project` would map to `/app/host/inertia/project`. Interestingly enough, we have to do some finnicking around that when we execute Docker commands - because the socket is on the host, we need to translate `/app/host/` paths back to `$HOME/` paths 😅 See https://github.com/ubclaunchpad/inertia/blob/v0.4.0-rc2/daemon/inertiad/project/build.go#L36

brian-nguyen: image

PR #274, opened 2018-06-24
Inertia web terminal view component refactor

### LINKED ISSUE #243: Inertia Web Terminal View Component
Take in an array of log messages, and produce the UI below:

- there should be a configurable fixed height; overflowed content should be scrollable; width should expand to its container (`width: 100%`)
- don't worry about making it an infinite-scroll component; we'll just pass in all the logs for now and add that feature later
- should be able to select and copy the text

screen shot 2018-06-13 at 10 33 58 pm

rogermyang:
:tickets: **Ticket(s)**: Closes #243

---

refactored and imporved sass style sheets; added null check to terminalview


daemon/web/components/TerminalView/TerminalView.js
```diff
@@ -1,26 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-
-const styles = {
```
bobheadxi: I noted in my other PR that this might be best to keep here, so as to allow programmatic styling via themes, but I suppose this is fine for now

daemon/web/components/TerminalView/TerminalView.js
```diff
@@ -1,26 +1,14 @@
...
-    resize: 'none',
-  },
-};
+import './index.sass';
 
 const TerminalView = props =>
   (
-    <div>
+    <div className="terminalView">
       <textarea
```
bobheadxi: Missed this before, I don't think `<textarea />` is valid

bobheadxi: actually my bad this is correct

PR #296, opened 2018-07-02
cmd: configuration rework

### LINKED ISSUE #101: config: improve format of inertia configuration
PR #99 introduces TOML configuration for Inertia. Being able to source control basic remote settings (IP, etc) might come in handy for some people, but there are certain things we probably don't want in there if someone chooses to commit the configuration file:

- Inertia daemon key: this one should definitely offer the option to be stored/read from elsewhere

- Local PEMfile location? Perhaps not critical, but a command line option to change the location might come in handy

An idea would be to allow remote config to be pulled from `ENV`

### LINKED ISSUE #350: Sent files get deleted upon deployment init
This will be resolved in #296

bobheadxi:
:tickets: **Ticket(s)**: Closes #101, closes #350

---

This PR splits configuration into two separate files:

* `inertia.toml`, which contains basic project configuration, and **should** be committed to source.

* `inertia.remotes`, which contains remote configuration, and **should not** be committed to source.

This opens the door for more project-specific configuration that can be source-controlled, as well as the ability to move configuration currently in `inertia.remotes` out of the project directory into a safer, centralized location to avoid accidental commits. For now, both will live in the user's project directory.

This PR also includes some miscellaneous tests.

```
$> make cli
$> inertia --project example.inertia.toml --remotes example.inertia.remotes production
```

## 🍧 Example

```toml
# inertia.toml
version = "latest"
project-name = "inertia"
build-type = "dockerfile"
build-file-path = "Dockerfile"
```

```toml
# inertia.remotes
version = "latest"

[remotes]
  [remotes.production]
    IP = "1.2.3.4"
    user = "root"
    pemfile = "/Users/robertlin/.ssh/id_rsa"
    branch = "master"
    ssh_port = "22"
    [remotes.production.daemon]
      port = "4303"
      token = "asdfasdf"
      webhook_secret = "9oC_5rfFcx2es4NLonMP2leOlJARkmu404EAXHXR4CI="
  [remotes.staging]
    IP = "4.3.2.1"
    user = "bleh"
    pemfile = "/Users/robertlin/.ssh/id_rsa"
    branch = "dev"
    ssh_port = "22"
    [remotes.staging.daemon]
      port = "4303"
      token = "asdfasdf"
      webhook_secret = "7dtg2lmTHF5EDBUgIpdu0zuERqs9Emn8rd3NCh5EiMU="
```


bobheadxi:
@chadlagore and I discussed this quite a bit way back in the day - one way would be to quietly slip in a `gitignore` rule for `inertia.remote`. Personally I've never seen a tool do this - though neither have I seen a tool output sensitive information into the project directory 😋

bobheadxi:
I could expand the scope of this PR to include a move of sensitive information into the user's home directory, or add something that adds a `gitignore` rule. I'm not a big fan of the latter. Any thoughts @PiggySpeed ?

PiggySpeed:
@bobheadxi  I'm more in favour of extending the task to move it to the user's home directory. Somewhere that the user can't possibly make a mistake committing the wrong file. With the names so similar: `inertia.toml` and `inertia.remotes`, it's a mistake waiting to happen.

leesw98:
My thoughts:
1. dot file > regular file: It will exclude the general audience (as in they can't see it naturally). At least, the default directory, where the inertia configs are stored, ie) `.inertia/`, should be dot'ed. Once inside, the config files themselves can be regular files.
2. I think the quietly slipping to a `.gitignore` is probably a bad idea as it causes silent side-effect.
3. Putting the config inside `$home` sounds good. If we do so, we should console print where the `inertia.remotes` is stored once the init is completed.
4. A disadvangtage would be that it would be hard to edit the `inertia.remotes` if they need to be edited often, but it seems like there is a command for it anyways.

bobheadxi:
Current plan:
- Remote configuration in `~/` (`HOME` on unix, `HOMEPATH` on windows), keyed by **file name** - this is `project`. There's some contention on whether this is ok, but AWS puts their config there, so I think it's safe for us to follow suit
- CLI options to specify either project name, or remote config path - cannot be set at the same time
- `inertia.toml` will be parsed by daemon and not the client - on first deploy, Inertia will use `inertia [remote] send` to send the config in the background to the deployment. This should keep the file there until the user commits `inertia.toml` to source control.

Usage patterns can look like this:

```bash
$ cd /any/directory
$ inertia --project bumper my_vps up
```

Thoughts, @PiggySpeed @leesw98 @brian-nguyen ?

bobheadxi:
Making some progress - still needs rather extensive testing to see if the new setup is a good experience. It's honestly a little convoluted tbh, but... oh well

Setup:

```bash
bobbook:inertia robertlin$ inertia remote add loc
Enter location of PEM file (leave blank to use '/Users/robertlin/.ssh/id_rsa'):

Enter user:
root
Enter IP address of remote:
127.0.0.1
Enter webhook secret (leave blank to generate one):

Enter project branch to deploy (leave blank for current branch):


Port 4303 will be used as the daemon port.
Port 22 will be used as the SSH port.
Run 'inertia remote add' with the -p flag to set a custom Daemon port
of the -ssh flag to set a custom SSH port.

Remote 'loc' has been added!
You can now run 'inertia loc init' to set this remote up
for continuous deployment.
bobbook:inertia robertlin$ cat ~/.inertia/inertia.remotes
version = "latest"

[remotes]
  [remotes.loc]
    IP = "127.0.0.1"
    user = "root"
    pemfile = "/Users/robertlin/.ssh/id_rsa"
    branch = "client/#101-config-rework"
    ssh-port = "22"
    [remotes.loc.daemon]
      port = "4303"
      token = ""
      webhook-secret = "fxX0qmgXipz2bih83OehGqNyYv3wQZd9YKOznFT5rsk="
```

cfg/config.go
```diff
@@ -2,45 +2,200 @@ package cfg
...
+	}
+	err = toml.Unmarshal(remotesBytes, remotes)
+	if err != nil {
+		return nil, fmt.Errorf("remotes config error: %s", err.Error())
+	}
+	return NewConfigFromTOML(*project, *remotes)
+}
+
+// NewConfigFromTOML loads configuration from TOML format structs
+func NewConfigFromTOML(project common.InertiaProject, remotes InertiaRemotes) (*Config, error) {
```
leesw98: Don't hate me for this but ... you are WRONG according to Microsoft and me. Lol.
Don't bother changing it though. It looks like you have been all capping acronyms everywhere. It would be a waste of time to change it now. Consistency > Preference
https://stackoverflow.com/questions/15526107/acronyms-in-camelcase

cfg/config.go
```diff
@@ -2,45 +2,200 @@ package cfg
...
+	// Set remote defaults
+	if remotes.Remotes == nil {
+		r := make(map[string]*RemoteVPS)
+		remotes.Remotes = r
```
leesw98: Is there a reason for doing these separately? It doesn't seem like `r` gets used elsewhere

cfg/config.go
```diff
@@ -2,45 +2,200 @@ package cfg
...
+	}
+	if remotes.Version == nil {
+		remotes.Version = project.Version
+	}
+
+	// Check all is g
+	if *project.Version != *remotes.Version {
+		return nil, fmt.Errorf("mismatching versions %s and %s", *project.Version, *remotes.Version)
```
leesw98: 👍

leesw98: Might be better to state what version was expected and what version was attempted.

cfg/config.go
```diff
@@ -2,45 +2,200 @@ package cfg
...
+			return remote, true
+		}
+	}
+	return nil, false
+}
+
+// AddRemote adds a remote to configuration
+func (config *Config) AddRemote(remote *RemoteVPS) bool {
+	_, ok := config.remotes[remote.Name]
+	if ok {
```
leesw98: Is this `ok` supposed to connote that the remote exists? Why not just call it `exists` or `isset` etc?

local/storage.go
```diff
@@ -5,18 +5,28 @@ import (
...
 
-	"github.com/BurntSushi/toml"
 	"github.com/ubclaunchpad/inertia/cfg"
 	"github.com/ubclaunchpad/inertia/client"
-	"github.com/ubclaunchpad/inertia/common"
 )
 
+// GetRemotesConfigFilePath retrieves path of file in global storage ($HOME/.inertia)
+func GetRemotesConfigFilePath(projectName string) string {
+	path := filepath.Join(".inertia", projectName+".remotes")
```
leesw98: 👍

local/storage.go
```diff
@@ -26,19 +36,14 @@ func InitializeInertiaProject(configPath, version, buildType, buildFilePath stri
...
-	if err != nil {
-		return err
-	}
-
+func createConfigFile(projectConfigPath, remoteConfigPath, version, buildType, buildFilePath, remoteURL string) error {
 	// Check if Inertia is already set up.
-	s, fileErr := os.Stat(configFilePath)
+	s, fileErr := os.Stat(projectConfigPath)
 	if s != nil {
 		return errors.New("inertia already properly configured in this folder")
```
leesw98: Probably should capitalize Inertia

common/util.go
```diff
@@ -80,6 +80,22 @@ func RemoveContents(directory string) error {
 	return nil
 }
 
+// isDirEmpty checks if given directory is empty
+func isDirEmpty(name string) (bool, error) {
+	f, err := os.Open(name)
+	if err != nil {
+		return false, err
+	}
+	defer f.Close()
```
leesw98: That's pretty cool.

cmd/deploy.go
```diff
@@ -137,18 +129,46 @@ This requires an Inertia daemon to be active on your remote - do this by running
...
+
+		if sendConfig {
+			println("reading project configuration...")
+			f, err := os.Open(projectConfigFilePath)
+			if err != nil {
+				log.Fatal(err.Error())
+			}
+
+			println("sending configuration to remote...")
+			projectPath := "$HOME/inertia/project"
```
leesw98: No love for Windows VPS 🙁

cfg/config.go
bobheadxi: Microsoft is wrong 😭

cfg/config.go
bobheadxi: I think here, no version is really "expected" - this just checks if the project config is the same inertia version as the remotes config. I think there's a separate check for CLI version vs config version somewhere that declares an expected version (though passing the check is not currently required)

cfg/config.go
bobheadxi: I think it's just a standard, though perhaps `found` would be more appropriate

common/util.go
bobheadxi: Definitely useful, though for arguably one of the most potential useful usecases (`defer mux.Unlock()`) the performance penalty is too significant to be of use

cmd/deploy.go
bobheadxi: Definitely needs windows tests - I've been hoping to set up Appveyor, but needs @bfbachmann to install it to the Launch Pad account (I can't seem to do it)

bobheadxi:
This PR just needs some final live testing - I've pushed up a `canary` image for this branch (21add01)

bobheadxi:
won't be merging this PR but eventually i rewrite this whole shebang using https://github.com/spf13/viper which now supports config writeback and clean up the mess of daemonside config initialization... leaving this open so i can reference it later tho

bobheadxi:
Closing this, in favour of #577

Issue #201, opened 2018-05-17
Pull, deploy, update Docker-image-based deployments


bobheadxi:
A great number of self-hosted services out there offer easily deployable Docker images. Would be handy to have Inertia be able to do things like:

1. `inertia init` in arbitrary directory
2. In config, set build type to `dockerhub` and add a Dockerhub image name to pull
3. When `inertia [REMOTE] up`, pull the appropriate image and deploy it

What this gives Inertia:

- users can now see a cool third-party project and effortlessly deploy it with Inertia
- users now have the option to use Dockerhub's automated builds, and register a webhook to notify Inertia of any updates. This will require some work on #133, however, to decouple our current webhook stuff from GitHub and make it more accepting of other webhooks.

PR #325, opened 2018-07-21
Webhook support for Bitbucket and Dockerhub

### LINKED ISSUE #307: Bitbucket webhook support
This is a followup to #133, which saw a lot of functionality added in PRs #229 and #301

brian-nguyen:
:tickets: **Ticket(s)**: Closes #307

---

Modified webhook package structure:
1. Unexported structs describing webhooks
2. Moved events to their own files.

Webhook support for Bitbucket Push event and Dockerhub added.
Docker's webhooks aren't as easily distinguishable and are different from the git based ones, so a dedicated handler and parser was implemented. Convenience methods to get relevant data are available.

Need some feedback!

Related: #201 

`make test`
Canary build is also available for live testing.


daemon/inertiad/webhook.go
```diff
@@ -34,6 +34,21 @@ func webhookHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// specialized handler for docker webhooks
+func dockerWebhookHandler(w http.ResponseWriter, r *http.Request) {
```
bobheadxi: any reason to make this separate from `webhookHandler`? I had imagined this would be part of that, though for now its fine for it to be separate (functionality not in place)

daemon/inertiad/webhook/bitbucket_push.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// Due to heavy nesting, extracting keys with type assertions is preferred
+func parseBitbucketPushEvent(rawJSON map[string]interface{}) bitbucketPushEvent {
+	// Extract push details
+	push := rawJSON["push"].(map[string]interface{})
+	changes := push["changes"].([]interface{})
+	changesObj := changes[0].(map[string]interface{})
+	new := changesObj["new"].(map[string]interface{})
+	branchName := new["name"].(string)
```
bobheadxi: 😮 this looks wild

daemon/inertiad/webhook/bitbucket_push.go
```diff
@@ -0,0 +1,59 @@
...
+// GetRef returns the full ref
+func (b bitbucketPushEvent) GetRef() string {
+	return "refs/heads/" + b.branchName
+}
+
+// GetGitURL returns the git clone URL
+// Ex. https://ubclaunchpad@bitbucket.org/ubclaunchpad/inertia.git
+func (b bitbucketPushEvent) GetGitURL() string {
+	user := strings.Split(b.fullName, "/")[0]
+	return "https://" + user + "@bitbucket.org/" + b.fullName + ".git"
```
bobheadxi: [`fmt.Sprintf`](https://golang.org/pkg/fmt/#Sprintf) is an option if you think the `+ ... +` is a bit messy 👌

daemon/inertiad/webhook/docker.go
```diff
@@ -0,0 +1,44 @@
+package webhook
+
+import "strings"
+
+// DockerWebhook represents a push to DockerHub
+// see https://docs.docker.com/docker-hub/webhooks/
+type DockerWebhook struct {
```
bobheadxi: 😍

daemon/inertiad/webhook.go
brian-nguyen: I had considered that and it'd be easy to make it a part of `webhookHandler` - json unmarshal would just fail if the body didn't match up, but then I realized that I wouldn't be able to satisfy the current `Payload` interface that we are returning. Make an even more generic interface?

bobheadxi: Oh you're right, hm... do you have any immediate ideas for making the interface more generic? If not, I think this is fine as is until we implement #201

brian-nguyen: Not really, but I didn't give it a lot of thought since like you mentioned, the Docker webhook functionality doesn't exist. I think this could be enough for now. In this PR, I've simply added the support for future functionality for Docker webhooks 😎

bobheadxi: Sounds good 😁

daemon/inertiad/webhook/bitbucket_push.go
brian-nguyen: The heavy nesting made it hard because I found myself making too many structs that contained only one property. If we do more heavy JSON parsing like this in the future, we can consider using a library that can abstract away those type assertions.

Examples:
https://github.com/antonholmquist/jason
https://github.com/buger/jsonparser

bobheadxi: I think this is way cleaner than defining the struct beforehand - let's go lib-less for now, love it!

brian-nguyen:
@bobheadxi Let me know what you think!

Tests are done.
Added the raw sample webhooks to their own test files.
Got rid of the intermediate structs that were there just for parsing JSON.

daemon/inertiad/webhook/bitbucket_push.go
```diff
@@ -42,14 +45,14 @@ func (b bitbucketPushEvent) GetRepoName() string {
 
 // GetRef returns the full ref
 func (b bitbucketPushEvent) GetRef() string {
-	return "refs/heads/" + b.branchName
+	return fmt.Sprintf("refs/heads/%s", b.branchName)
```
bobheadxi: 👍

daemon/inertiad/webhook/webhook_test.go
```diff
@@ -24,13 +22,13 @@ func TestParse(t *testing.T) {
 		eventHeader string
 		eventValue  string
 	}{
-		{githubPushBody, "x-github-event", GithubPushHeader},
-		{gitlabPushBody, "x-gitlab-event", GitlabPushHeader},
-		{bitbucketPushBody, "x-event-key", BitbucketPushHeader},
+		{githubPushRawJSON, "x-github-event", GithubPushHeader},
+		{gitlabPushRawJSON, "x-gitlab-event", GitlabPushHeader},
+		{bitbucketPushRawJSON, "x-event-key", BitbucketPushHeader},
```
bobheadxi: 💪

Still trying to convert the our tests to this format, didn't know about this standard until a month ago or so 😢

Issue #315, opened 2018-07-11
Create flag fetch helper


bobheadxi:
change:

```go
		admin, err := cmd.Flags().GetBool("admin")
		if err != nil {
			log.Fatal(err)
		}
```

to:

```go
		admin, _ := cmd.Flags().GetBool("admin")
```

it doesnt matter when flag fetches fail

bobheadxi:
Wait is this really a good idea

PR #355, opened 2018-08-03
Add option to load AWS credentials from file

### LINKED ISSUE #354: Allow use of aws credentials for ec2 provisioning
https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html

bobheadxi:
:tickets: **Ticket(s)**: Closes #354

---

New:

```
      --profile-path string   path to aws credentials file
      --profile-user string   user profile for aws credentials file
```

create a config file:

```
bobbook:inertia robertlin$ cat ~/.aws/credentials
[default]
aws_access_key_id = asdf
aws_secret_access_key = asdf
```

go!

```bash
$> inertia provision ec2 nw-dev --profile-path ~/.aws/credentials
```


provision/ec2_test.go
```diff
@@ -15,3 +15,7 @@ func TestNewEC2ProvisionerFromEnv(t *testing.T) {
 	prov, _ := NewEC2Provisioner("id", "key")
 	assert.NotNil(t, prov.client.Config.Credentials)
 }
+
+func TestNewEC2ProvisionerFromProfile(t *testing.T) {
+	_, _ = NewEC2ProvisionerFromProfile("", "")
```
leesw98: Great test

cmd/provision.go
```diff
@@ -60,26 +66,35 @@ This ensures that your project ports are properly exposed and externally accessi
...
+		} else if profilePath != "" {
+			profileUser, _ := cmd.Flags().GetString("profile-user")
+			prov, err = provision.NewEC2ProvisionerFromProfile(profilePath, profileUser, os.Stdout)
 			if err != nil {
 				log.Fatal(err)
 			}
 		} else {
-			prov, err = provision.NewEC2ProvisionerFromEnv(os.Stdout)
+			var id, key string
+			id, key, err = enterEC2CredentialsWalkthrough(os.Stdin)
```
leesw98: Is there a particular reason why you separated variable declaration? I thought you can do things like `id, key, err := ...`

cmd/provision.go
```diff
@@ -60,26 +66,35 @@ This ensures that your project ports are properly exposed and externally accessi
...
 
-		// Load flags
+		// Load flags for credentials
 		fromEnv, _ := cmd.Flags().GetBool("from-env")
-		instanceType, _ := cmd.Flags().GetString("type")
+		profilePath, _ := cmd.Flags().GetString("profile-path")
+
+		// Load flags for setup configuration
 		user, _ := cmd.Flags().GetString("user")
+		instanceType, _ := cmd.Flags().GetString("type")
```
leesw98: I was gonna complain about these flag getters are really repeative, but I couldn't find a better solution. Do you know any?

cmd/provision.go
```diff
@@ -60,26 +66,35 @@ This ensures that your project ports are properly exposed and externally accessi
...
-			id, key, err = enterEC2CredentialsWalkthrough(os.Stdin)
+		if fromEnv {
+			prov, err = provision.NewEC2ProvisionerFromEnv(os.Stdout)
 			if err != nil {
 				log.Fatal(err)
 			}
-			prov, err = provision.NewEC2Provisioner(id, key, os.Stdout)
+		} else if profilePath != "" {
+			profileUser, _ := cmd.Flags().GetString("profile-user")
+			prov, err = provision.NewEC2ProvisionerFromProfile(profilePath, profileUser, os.Stdout)
```
leesw98: Note for self: Test this part later (the new logic).

provision/ec2_test.go
bobheadxi: Thanks - I think testing this properly is a bit of a stretch goal for now, since the API is a pain in the ass

cmd/provision.go
bobheadxi: Good point, I think this was a holdover from something else I had intended to do, where `id, key` could be set from different functions depending on the condition. I'll simplify it

cmd/provision.go
bobheadxi: The oneline is required for sure, and I think it's fine - the err check is meh.

I thought about it again, and I realized we **can** ignore the err check probably - broken inputs will return a 0-value (for strings, it's `""`) which is safe. Doh. Maybe this calls for a reopening of #315? what do you think?

cmd/provision.go
bobheadxi: I used this last night, to test Inertia builds for nwhacks, but feel free to give it a try!

Issue #293, opened 2018-07-01
Quick rollback


bobheadxi:
Keep images (or containers?) built for the last few project versions (say, 3) to allow easy rollbacks. Add `inertia [remote] rollback ...` commands.

PR #356, opened 2018-08-03
Dockerfile ports, containerstop fixes, minor updates

### LINKED ISSUE #351: Dockerfile builds don't expose ports properly

bobheadxi:
:tickets: **Ticket(s)**: Closes #351

---

See commits - biggest change is that ports exposed in Dockerfile images (via the `EXPOSE` directive) is now automatically mapped to host

build any dockerfile project, check that port is accessible

### DIFF
--- a/daemon/inertiad/build/builder.go
+++ b/daemon/inertiad/build/builder.go
@@ -12,6 +12,7 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	docker "github.com/docker/docker/client"
+	"github.com/docker/go-connections/nat"
 	"github.com/ubclaunchpad/inertia/daemon/inertiad/cfg"
 	"github.com/ubclaunchpad/inertia/daemon/inertiad/containers"
 	"github.com/ubclaunchpad/inertia/daemon/inertiad/log"
@@ -151,7 +152,7 @@ func (b *Builder) dockerCompose(d Config, cli *docker.Client,
 	}
 
 	// Start container to build project
-	fmt.Fprintln(out, "Building project...")
+	fmt.Fprintf(out, "Building project %s...\n", d.Name)
 	err = cli.ContainerStart(ctx, resp.ID, types.ContainerStartOptions{})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


daemon/inertiad/build/builder.go
```diff
@@ -299,7 +312,7 @@ func (b *Builder) herokuishBuild(d Config, cli *docker.Client,
 	}
 
 	// Start the herokuish container to build project
-	fmt.Fprintln(out, "Building project...")
+	fmt.Fprintf(out, "Building project %s...\n", d.Name)
```
leesw98: Perhaps you'd want to put duplicate strings in a var for consistency

daemon/inertiad/build/builder.go
```diff
@@ -243,16 +247,25 @@ func (b *Builder) dockerBuild(d Config, cli *docker.Client,
...
 	fmt.Fprintf(out, "%s (%s) build has exited\n", imageName, buildResp.OSType)
 
+	// Get image details
+	image, _, err := cli.ImageInspectWithRaw(ctx, imageName)
+	if err != nil {
+		return nil, err
+	}
+	portMap := nat.PortMap{}
+	for p := range image.Config.ExposedPorts {
+		portMap[p] = []nat.PortBinding{{HostIP: "0.0.0.0", HostPort: p.Port()}}
```
leesw98: `{{...}}` wot. Is hash parameters not a thing in Go?

daemon/inertiad/containers/containers.go
```diff
@@ -94,10 +94,13 @@ func StopActiveContainers(docker *docker.Client, out io.Writer) error {
...
 			timeout := 10 * time.Second
-			err := docker.ContainerStop(ctx, container.ID, &timeout)
-			if err != nil {
+			if err := docker.ContainerStop(ctx, container.ID, &timeout); err != nil {
 				return err
 			}
+
+			// Archive container
+			docker.ContainerRename(
+				ctx, container.ID, fmt.Sprintf("%s-%d", container.Names[0], time.Now().Unix()))
```
leesw98: What's the point of renaming these containers? Wouldn't we pile up on archived containers in this case if ran for long period?

daemon/inertiad/build/builder.go
bobheadxi: What do you mean?

The outer `{}` is the array instantiator, the inner `{}` is for an element

daemon/inertiad/build/builder.go
bobheadxi: I was considering that, but wasn't sure if it would be within the scope of the PR - though considering theres a bunch of misc changes, might as well

daemon/inertiad/containers/containers.go
bobheadxi: Ran into some issues with duplicate container names in my testing. The reasoning here is twofold:
- unused containers will be removed by our `Prune()` op, which runs at certain points (pre-build, etc)
- archival might be useful for #293

We definitely need a stronger plan for resource management, but I still have some confusions about how the fuck docker itself manages it. It seems like there is a delay in API call -> thing actually happening, as can be seen in the Travis tests periodically failing. Definitely something that needs more research.

PR #357, opened 2018-08-03
Add newlines to ec2 provision output and refactor logic

### LINKED ISSUE #348: Provision output needs newlines
```Creating t2.micro instance in us-east-1 from image ami-f316478c...
Generating key pair nwhacks-test_ec2-user_inertia_key_1533265801135866606...
Saving key to /Users/robertlin/.ssh/nwhacks-test_ec2-user_inertia_key_1533265801135866606...
Checking status of requested instance...Instance status: pendingInstance status: pendingInstance status: pendingInstance status: pendingIn```

### PR DESCRIPTION
**Pull request by bobheadxi**
:tickets: **Ticket(s)**: Closes #348

---

Add newlines back to output
Improve flow and error checking

```bash
$> inertia provision ec2 new-vps
```


provision/ec2.go
```diff
@@ -165,39 +166,70 @@ func (p *EC2Provisioner) CreateInstance(opts EC2CreateInstanceOptions) (*cfg.Rem
...
-	var instanceStatus *ec2.DescribeInstancesOutput
+	fmt.Fprintln(p.out, "Checking status of requested instance...")
+	var (
+		attempts = 0
+		instance ec2.Instance
+	)
 	for {
+		// Wait briefly between checks
+		time.Sleep(3 * time.Second)
 		attempts++
```
leesw98: This seems to be a unused variable. Perhaps you wanted to increase the check time period?

provision/ec2.go
```diff
@@ -165,39 +166,70 @@ func (p *EC2Provisioner) CreateInstance(opts EC2CreateInstanceOptions) (*cfg.Rem
...
 			InstanceIds: []*string{runResp.Instances[0].InstanceId},
 		})
 		if err != nil {
 			return nil, err
 		}
-		if len(result.Reservations) == 0 || len(result.Reservations[0].Instances) == 0 {
+
+		// Check if reservations are present
+		if result.Reservations == nil || len(result.Reservations) == 0 ||
+			len(result.Reservations[0].Instances) == 0 {
```
leesw98: Is there something similar to ruby's [`#dig`](https://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig)? That would simplify this a lot. I understand this is a struct but thought it could be nice to have since you are refactoring.

provision/ec2.go
```diff
@@ -165,39 +166,70 @@ func (p *EC2Provisioner) CreateInstance(opts EC2CreateInstanceOptions) (*cfg.Rem
...
+			fmt.Println(p.out, "Status unknown.")
 			continue
-		} else if *result.Reservations[0].Instances[0].State.Code == 16 {
-			// Code 16 means we can continue!
-			fmt.Fprint(p.out, "Instance is running!")
-			instanceStatus = result
+		}
+
+		// Code 16 means instance has started, and we can continue!
+		if s.Code != nil && *s.Code == 16 {
```
leesw98: We could make a global variable called `EC2_INSTANCE_STARTED_CODE` or something. Then we'd have self-documenting code! 😄

provision/ec2.go
```diff
@@ -210,17 +242,17 @@ func (p *EC2Provisioner) CreateInstance(opts EC2CreateInstanceOptions) (*cfg.Rem
 		},
 	})
 	if err != nil {
-		fmt.Fprintf(p.out, "Failed to set tags: %s", err.Error())
+		fmt.Fprintln(p.out, "Failed to set tags: "+err.Error())
```
leesw98: Wait I think the variable passing into string is better than direct concat. Does `fmt.Fprintln` not allow it or something?

provision/ec2.go
bobheadxi: Why global? The code is part of the API

provision/ec2.go
bobheadxi: The idiomatic way to do this in Go is to provide `Get()` functions, which Amazon has been kind enough to not provide. I think there's a dereference helper function coming in #296, but I'd rather just do it in one go if we do refactor this into a less verbose check - I think I'll leave it as it is for the time being

provision/ec2.go
bobheadxi: Right, I think this was supposed to be a cap on the number of attempts. I might remove it entirely.

provision/ec2.go
bobheadxi: Unfortunately there's no `fmt.Fprintfln()` (probably for good reason, because thats a ridiculous function name)

I think it's because for here, I've decided to standardize on using concat if the attached variable is the last part of a string - there's a few example of the code where i use formatting directives to insert strings into the middle

provision/ec2.go
bobheadxi: This paradigm is interesting to: https://github.com/storozhukBM/verifier - thoughts?

provision/ec2.go
bobheadxi: I feel like the AWS library should be the one that declares this. Oh well

bobheadxi: image

PR #364, opened 2018-08-05
Web: new eslint, minor refactors

### LINKED ISSUE #336: Update web linter
eslint5

bobheadxi:
:tickets: **Ticket(s)**: Closes #336

---

- Updated web code for `eslint5` and its new rules
- Updated stylesheets to match new `sass-lint`
- Refactored stylesheets where appropriate to simplify things
- Refactored components, pages so that they use `index.js` to avoid deeply nested import paths:

```js
import Status from '../../components/status/Status';
```

becomes:

```js
import Status from '../../components/Status';
```

```
make web-run
```

tbh I still need to test this, but opening it up for critique


daemon/web/index.html
```diff
@@ -9,7 +9,7 @@
 </head>
 
 <body>
-  <div id="app"></div>
+  <div id="app" class="app"></div>
```
PiggySpeed: Could we remove the need for `class="app"`? and use the id instead?

daemon/web/pages/Dashboard/index.js
```diff
@@ -32,18 +32,21 @@ const styles = {
 
 class DashboardWrapper extends React.Component {
   componentWillMount() {
-    this.props.handleGetProjectDetails();
-    this.props.handleGetContainers();
+    const { handleGetProjectDetails, handleGetContainers } = this.props;
```
PiggySpeed: Not sure why this is necessary - we're adding two extra assignments?

bobheadxi: The new eslint rules are more strict about destructuring, and I think occasionally having situations like this is a small price to pay to avoid this.props.thing.another.thing that’s already a bit prevalent

daemon/web/index.html
bobheadxi: Sass-lint prohibits use of ID selectors, and since I could only find one instance of its usage I opted to just change this to a class instead rather than disable the rule.

PiggySpeed: hmm... ok....

bobheadxi: I suppose it's to discourage use of styles that only apply to a single element, which kind of makes sense

bobheadxi:
As a note - I can't seem to repro the issue that travis is having, the web app builds fine on my machine 🙁

bobheadxi:
what the fuck it was because the full path of `Containers/index.sass` was lowercased in git, but not in my filesystem...

PR #398, opened 2018-08-27
Daemon/#156 generate token

### LINKED ISSUE #156: Generate 'master' tokens
Generate tokens associated with permission levels for team leads to share - something like:

```
inertia $VPS_NAME user token --admin
```

For example, an `admin` token should be allowed to add users to Inertia Web and generate new tokens.

This will require the changes from #139 first, which introduces the `inertia $VPS_NAME user` commands

Should also make sure that Inertia (outside of `inertia $VPS_NAME init` and `ssh`) can be fully functional without the public key.

leesw98:
:tickets: **Ticket(s)**: Closes #156

---

Added a new command for generating admin tokens

Test that `inertia [remote] token` command generates a new token.

Note: @bobheadxi feel free to comment about text changes.


client/client.go
```diff
@@ -245,6 +245,11 @@ func (c *Client) Up(gitRemoteURL, buildType string, stream bool) (*http.Response
 	})
 }
 
+// Token generates token on this remote.
+func (c *Client) Token() (*http.Response, error) {
```
bobheadxi: could we get a useless test here to inflate our coverage

daemon/inertiad/token.go
```diff
@@ -0,0 +1,26 @@
+package main
+
+import (
+	"net/http"
+
+	"github.com/ubclaunchpad/inertia/daemon/inertiad/crypto"
+)
+
+// tokenHandler generates a new token
+func tokenHandler(w http.ResponseWriter, r *http.Request) {
```
bobheadxi: same here, this one actually isn't that useless to test. refer to other handler for how you might go about setting this one up for tests

daemon/inertiad/token.go
```diff
@@ -0,0 +1,26 @@
...
+	keyBytes, err := crypto.GetAPIPrivateKey(nil)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
suujia: `internalServerError` instead?

daemon/inertiad/token.go
```diff
@@ -0,0 +1,26 @@
...
+		return
+	}
+
+	token, err := crypto.GenerateMasterToken(keyBytes.([]byte))
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
suujia: same here, this isn't anything wrong with the request

daemon/inertiad/token.go
leesw98: Yup. Thanks 👍

PR #401, opened 2018-08-30
Logout button on Nav

asitu:
:tickets: **Ticket(s)**: Closes #

---

Logout button on the nav (as a component) without global styling

### DIFF
--- a/daemon/web/components/buttons/LogoutButton/index.js
+++ b/daemon/web/components/buttons/LogoutButton/index.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import '../index.sass';
+
+const LogoutButton = ({ style }) => (
+  <div>
+    <button className="button" type="button" style={style}>
+Logout
+    </button>
+  </div>
+);
+
+LogoutButton.propTypes = {
+  style: PropTypes.object,
+};
+
+export default LogoutButton;
--- a/daemon/web/components/buttons/LogoutButton/index.js
+++ b/daemon/web/components/buttons/LogoutButton/index.js
@@ -1,18 +1,13 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 
 import '../index.sass';
 
-const LogoutButton = ({ style }) => (
+const LogoutButton = () => (
   <div>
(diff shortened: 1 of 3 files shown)


daemon/web/components/buttons/LogoutButton/index.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import '../index.sass';
+
+const LogoutButton = ({ style }) => (
```
bobheadxi: LogoutButton should always look the same, right? I think we can omit the `style` prop

daemon/web/components/buttons/ShutdownButton/index.js
```diff
@@ -1,18 +1,18 @@
...
+const ShutdownButton = () => (
   <div>
-    <button className="button" type="button" style={style}>
+    <button className="button" type="button">
 Shut Down Containers
     </button>
   </div>
 );
 
 ShutdownButton.propTypes = {
```
bobheadxi: since there are no props, this can be removed entirely :)

bobheadxi: WAIT i think you changed the wrong button? this is `ShutdownButton`

bobheadxi: should probably be updating `LogoutButton` instead

asitu: Okay lol yep.. Should it be removed for all the buttons? all the buttons now don't follow the global style classes

bobheadxi: I think we should leave it for the other buttons for now, until we can check if they use it or not.

The buttons as a whole probably need a rework - we probably don't want `ShutdownButton` and `LogoutButton`, but a `PrimaryButton` and `SecondaryButton` instead, to improve re-usability - but this is fine for the time being.

bobheadxi: So yeah, just remove it for `LogoutButton`, and leave it as is for the others 🚀

asitu: OKayy 👍

PR #423, opened 2018-10-25
[WIP] Allow clients to authenticate via the CLI

### LINKED ISSUE #421: Update CLI to allow regular user login

bfbachmann:
:tickets: **Ticket(s)**: Closes #421 

---

@bobheadxi I marked this as WIP because I'm not sure we can actually merge this yet. With the functionality I've added, if you log in as a regular user while you hold the master token, that master token will be overwritten with the auth token for the user you logged in as. Is there a way to add regular credentials to the master account as-is, or will we need to implement that? 

- Added `LogIn` method to `users.go` that hits the `/user/auth` endpoint with the user's credentials and returns the response.
- Added tests for `LogIn()`.
- Added a `login` command that accepts the user's credentials (password is hidden and can't be specified as an argument) and uses them to retrieve an auth token. The `inertia.toml` config file is then updated with the username and token if the authentication succeeds.

Just run the tests, or use the login command as a regular, non-master user.

### DIFF
--- a/cmd/users.go
+++ b/cmd/users.go
@@ -108,6 +108,58 @@ from the web app.`,
 	},
 }
 
+var cmdDeploymentLogin = &cobra.Command{
+	Use:   "login",
+	Short: "Authenticate with the remote",
+	Long:  "Retreives an access token from the remote using your credentials.",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		remoteName := strings.Split(cmd.Parent().Parent().Use, " ")[0]
+		deployment, _, err := local.GetClient(remoteName, configFilePath, cmd)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		username := args[0]
+		fmt.Print("Password: ")
+		pwBytes, err := terminal.ReadPassword(int(syscall.Stdin))
+		fmt.Println()
+		if err != nil {
+			log.Fatal(err)
+		}
+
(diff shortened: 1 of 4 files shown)


bobheadxi:
@bfbachmann updating this branch with the changes in master will allow CI to pass

bobheadxi:
> With the functionality I've added, if you log in as a regular user while you hold the master token, that master token will be overwritten with the auth token for the user you logged in as.

This is fine I think

> Is there a way to add regular credentials to the master account as-is, or will we need to implement that?

No such functionality at the moment unfortunately, though it would be nice, though the best way to do this would probably to remove the use of a `master` user and instead use a separate permissions field in the token, and allow the use of the token generation feature from #398 to create non-expiring API tokens

cmd/users.go
```diff
@@ -108,6 +108,58 @@ from the web app.`,
...
+		fmt.Print("Password: ")
+		pwBytes, err := terminal.ReadPassword(int(syscall.Stdin))
+		fmt.Println()
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		resp, err := deployment.LogIn(username, string(pwBytes))
+		if err != nil {
+			log.Fatal(err)
```
bobheadxi: we might want slightly better feedback here, in case daemon is offline and whatnot

bfbachmann: Good call. I'll update this.

bobheadxi:
@bfbachmann can I merge this?

PR #436, opened 2018-11-02
Totp/#417 crypto

### LINKED ISSUE #417: Add code for doing OTP-related crypto to `crypto` module
We need support for
- AES (I think we already have this, but check)
- TOTP
- PDKDF2

tristancalderbank:
:tickets: **Ticket(s)**: Closes #417 

---

- Adds symmetric encryption/decryption using AES in GCM mode
- Adds password-based key derivation using PBKDF2 via HMAC-SHA256

See unit tests


daemon/inertiad/crypto/key_derivation.go
```diff
@@ -0,0 +1,34 @@
...
+
+// DeriveKey derives an AES encryption key based on salt + user's
+// password using PBKDF2 with HMAC-SHA1
+func DeriveKey(password string, salt []byte) []byte {
+	return pbkdf2.Key(
+		[]byte(password),
+		salt,
+		keyDerivationIterations,
+		KeyDerivationKeyLength,
+		sha1.New,
```
bfbachmann: _Please_ can we not use SHA1.

daemon/inertiad/crypto/symmetric.go
```diff
@@ -0,0 +1,63 @@
...
+	return nonce
+}
+
+// Encrypt encrypts plaintext using given key in AES GCM mode
+func Encrypt(key, plaintext []byte) ([]byte, error) {
+	block, err := aes.NewCipher(key)
+
+	if err != nil {
+		return nil, err
+	}
```
bfbachmann: Minor: In Go you usually do
```go
something, err := expr()
if err != nil {
...
```

In other words, you generally want no whitespace between the call that returns the error and the error check so as to create clear groupings of error handling logic in your code.

daemon/inertiad/crypto/key_derivation.go
tristancalderbank: Yeah we can use SHA-256 if we want.

But FYI there's no weakness in SHA-1 when it's used for HMAC:

https://stackoverflow.com/a/4944822

Whats the issue?

daemon/inertiad/crypto/symmetric.go
tristancalderbank: fixed

daemon/inertiad/crypto/key_derivation.go
tristancalderbank: should be fixed

tristancalderbank:
Yeah if we get time we could replace it

tristancalderbank:
Right, this should be the end of this branch, has all we need as far as crypto for 2FA

bfbachmann:
@tristancalderbank Looks awesome! Thanks.

PR #461, opened 2018-11-23
Update distable totp

bfbachmann:
- Fixed a bug in the client implementation of `disableTotp` (it should take no arguments because the user is authenticated user his/her access token)
- Fixed some minor code smells

### DIFF
--- a/cmd/users.go
+++ b/cmd/users.go
@@ -151,6 +151,7 @@ var cmdDeploymentLogin = &cobra.Command{
 			}
 		}
 
+		fmt.Printf("(Status code %d) ", resp.StatusCode)
 		if resp.StatusCode != http.StatusOK {
 			fmt.Println("Invalid credentials")
 			return
@@ -274,7 +275,7 @@ var cmdDeploymentEnableTotp = &cobra.Command{
 		}
 
 		if resp.StatusCode != http.StatusOK {
-			fmt.Printf("(Status code %d) Error Enabling Totp.\n", resp.StatusCode)
+			fmt.Printf("(Status code %d) Error Enabling Totp.", resp.StatusCode)
 			return
 		}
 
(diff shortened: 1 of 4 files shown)


cmd/users.go
```diff
@@ -287,25 +288,25 @@ var cmdDeploymentEnableTotp = &cobra.Command{
...
 		if err = json.Unmarshal(body, &totpInfo); err != nil {
 			log.Fatal(err)
-			return
 		}
 
-		qrBuilder := qr.New()
-		qrBuilder.Get(fmt.Sprintf("otpauth://totp/%s?secret=%s&issuer=Inertia",
+		// Display QR code so users can easily add their keys to their
+		// authenticator apps
+		qr.New().Get(fmt.Sprintf("otpauth://totp/%s?secret=%s&issuer=Inertia",
```
bobheadxi: 😍

bfbachmann: Yea... amazingly this actually works with Google Authenticator. It probably works with other apps too.

daemon/inertiad/auth/permissions.go
```diff
@@ -13,8 +14,12 @@ import (
 	"github.com/ubclaunchpad/inertia/daemon/inertiad/util"
 )
 
+// ctxKey represents keys used in request contexts
+type ctxKey int
+
 const (
-	errMalformedHeaderMsg = "malformed authorization error"
+	errMalformedHeaderMsg        = "malformed authorization error"
+	ctxUsername           ctxKey = iota
```
bobheadxi: never worked with `iota` before, do you mind explaining how this is used?

cmd/users.go
bobheadxi: thats sooooo slick

daemon/inertiad/auth/permissions.go
bfbachmann: No problem. `iota` just assigns consecutive integers to constants (docs [here](https://github.com/golang/go/wiki/Iota)). The reason I'm doing this is because Go's `Context` requires that keys with which values are associated are their own special type (they can't just be strings like in a regular `map`). This assignment is basically just setting `ctxUsername = 0`.

bobheadxi: ahh i see, thanks!

daemon/inertiad/auth/users.go
```diff
@@ -3,14 +3,15 @@ package auth
...
+
 	"github.com/boltdb/bolt"
 	"github.com/ubclaunchpad/inertia/daemon/inertiad/crypto"
 )
 
 var (
 	errSessionNotFound    = errors.New("Session not found")
 	errUserNotFound       = errors.New("User not found")
-	errBackupCodeNotFound = errors.New("backup code not found")
+	errBackupCodeNotFound = errors.New("Backup code not found")
```
tristancalderbank: I'll make these lowercase

PR #552, opened 2019-02-12
cmd: add cross-platform GetHomePath()

### LINKED ISSUE #400: provision/ec2: windows compatibility
The following:

https://github.com/ubclaunchpad/inertia/blob/master/provision/ec2.go#L140

```go
	// Save key
	keyPath := filepath.Join(os.Getenv("HOME"), ".ssh", *keyResp.KeyName)
	fmt.Printf("Saving key to %s...\n", keyPath)
	err = local.SaveKey(*keyResp.KeyMaterial, keyPath)
	if err != nil {
		return nil, err
	}
```

probably doesn't work - [HOME is not usually set in Windows](https://superuser.com/questions/607105/is-the-home-environment-variable-normally-set-in-windows)

blocked by #296, which will land with improved Windows compatibility for these things on the config side

yaoharry:
:tickets: **Ticket(s)**: for #400  and #519 

---
Added an operating system check for setting windows home path.

null check that home path was successfully set

### DIFF
--- a/local/env.go
+++ b/local/env.go
@@ -25,9 +25,9 @@ func GetHomePath() (string, error) {
 		}
 
 		// Builds HOME from HOMEDRIVE and HOMEPATH as default
-		drive := os.Getenv("HOMEDRIVE")
+		var drive = os.Getenv("HOMEDRIVE")
 		var path = os.Getenv("HOMEPATH")
-		home := drive + path
+		var home = drive + path
 		if drive == "" || path == "" {
 			return "", errors.New("HOMEDRIVE, HOMEPATH, or USERPROFILE environment variables are blank")
 		}
--- a/local/env.go
+++ b/local/env.go
@@ -1,6 +1,42 @@
 package local
 
+import (
+	"errors"
+	"os"
+	"runtime"
+)
+
 const (
 	// EnvSSHPassphrase is the key used to fetch PEM key passphrases
 	EnvSSHPassphrase = "PEM_PASSPHRASE"
 )
+
+// GetHomePath returns the homepath based on the operating system used
+func GetHomePath() (string, error) {
+	if runtime.GOOS == "windows" {
+		// Performs check for HOME env variable first
+		if home := os.Getenv("HOME"); home != "" {
+			return home, nil
+		}
+
+		// Performs check for USERPROFILE as backup
+		if home := os.Getenv("USERPROFILE"); home != "" {
+			return home, nil
+		}
+
+		// Builds HOME from HOMEDRIVE and HOMEPATH as default
+		drive := os.Getenv("HOMEDRIVE")
+		path := os.Getenv("HOMEPATH")
+		home := drive + path
+		if drive == "" || path == "" {
+			return "", errors.New("HOMEDRIVE, HOMEPATH, or USERPROFILE environment variables are blank")
+		}
+
+		return home, nil
+	}
+	// Unix system as default
+	if home := os.Getenv("HOME"); home != "" {
+		return home, nil
+	}
+	return "", errors.New("HOME environment variable is blank")
+}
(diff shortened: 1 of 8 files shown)


yaoharry:
@bobheadxi Just saw your issue on converting assert nil checks to NoError checks. Let's discuss Saturday on how to write better unit tests.

provision/ec2_test.go
```diff
@@ -23,3 +23,8 @@ func TestNewEC2ProvisionerFromProfile(t *testing.T) {
 	assert.NotNil(t, prov.client.Config.Credentials)
 	assert.Equal(t, "bob", prov.GetUser())
 }
+
+func TestHomePath(t *testing.T) {
+	homePath := GetHomePath()
+	assert.NotNil(t, homePath)
```
bobheadxi: `GetHomePath()` returns a string, so the assertion here might be [NotEqual](https://godoc.org/github.com/stretchr/testify/assert#NotEqual) empty string

provision/ec2.go
```diff
@@ -57,6 +58,14 @@ func NewEC2ProvisionerFromProfile(user, profile, path string, out ...io.Writer)
 // GetUser returns the user attached to given credentials
 func (p *EC2Provisioner) GetUser() string { return p.user }
 
+// GetHomePath returns the homepath based on the operating system used
+func GetHomePath() string {
+	if runtime.GOOS == "windows" {
```
bobheadxi: 🤔 this is neat

bobheadxi:
@yaoharry yep, thats a good point - the codebase is quite old and there are quite a lot of examples of things *not* to do 😅 remind me this saturday!

for the time being I think the test you have here is adequate. One thing we might want to consider is whether `$HOMEPATH/.ssh` is an appropriate directory for storing SSH keys in Windows - if you have some time this week do you mind looking into that?

provision/ec2.go
```diff
@@ -57,6 +58,14 @@ func NewEC2ProvisionerFromProfile(user, profile, path string, out ...io.Writer)
 // GetUser returns the user attached to given credentials
 func (p *EC2Provisioner) GetUser() string { return p.user }
 
+// GetHomePath returns the homepath based on the operating system used
+func GetHomePath() string {
```
bobheadxi: oh and we might want to move this function to package `local` instead, since it pertains to local filesystem and might come in handy in the future

https://github.com/ubclaunchpad/inertia/tree/master/local

yaoharry: Yeah good point. I was wondering if go had something for that. Would definitely help with the home paths in other windows bug fixes.

yaoharry: HomePath function has been migrated to local folder. Need the commit merged before I can import the function in the EC2 homepath fetching.

local/env.go
```diff
@@ -1,6 +1,42 @@
...
+		}
+
+		// Performs check for USERPROFILE as backup
+		if home := os.Getenv("USERPROFILE"); home != "" {
+			return home, nil
+		}
+
+		// Builds HOME from HOMEDRIVE and HOMEPATH as default
+		drive := os.Getenv("HOMEDRIVE")
+		path := os.Getenv("HOMEPATH")
```
bobheadxi: nitpicking, but for standalone, single-variable assignments I prefer the use of `var` for clarity (ie make it clear that a new variable is assigned created here)

```suggestion
		var path = os.Getenv("HOMEPATH")
```

not a big deal though!

local/env_test.go
```diff
@@ -0,0 +1,14 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetHomePath(t *testing.T) {
+	env, err := GetHomePath()
+	assert.NoError(t, err)
+	assert.NotEqual(t, env, "")
+	assert.DirExists(t, env)
```
bobheadxi: definitely a good thing to check 👍

Issue #407, opened 2018-09-27
Azure host support


bobheadxi:
Docker socket not available on windows

https://github.com/MicrosoftDocs/Virtualization-Documentation/issues/442

PR #568, opened 2019-02-15
Switched home fetch to local function with windows compatability

### LINKED ISSUE #400: provision/ec2: windows compatibility

### LINKED ISSUE #519: CLI Windows compatibility

yaoharry:
:tickets: **Ticket(s)**: Closes #400 , closes #519 

---

Updated home path checking with local function with windows compatibility check

- Ensure `$HOME` environment variable is the same as returned value from `local.GetHomePath()`
- Run main unit tests

### DIFF
--- a/cmd/inpututil/input.go
+++ b/cmd/inpututil/input.go
@@ -4,8 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"os"
 	"path/filepath"
+
+	"github.com/ubclaunchpad/inertia/local"
 
 	"github.com/ubclaunchpad/inertia/cfg"
 	"github.com/ubclaunchpad/inertia/common"
@@ -26,13 +27,16 @@ func AddRemoteWalkthrough(
 	in io.Reader, config *cfg.Config,
 	name, port, sshPort, currBranch string,
 ) error {
-	homeEnvVar := os.Getenv("HOME")
+	homeEnvVar, err := local.GetHomePath()
+	if err != nil {
+		return err
+	}
 	sshDir := filepath.Join(homeEnvVar, ".ssh")
 	defaultSSHLoc := filepath.Join(sshDir, "id_rsa")
 
 	var response string
 	fmt.Println("Enter location of PEM file (leave blank to use '" + defaultSSHLoc + "'):")
-	_, err := fmt.Fscanln(in, &response)
+	_, err = fmt.Fscanln(in, &response)
 	if err != nil {
 		response = defaultSSHLoc
 	}
--- a/cmd/inpututil/input.go
+++ b/cmd/inpututil/input.go
@@ -6,10 +6,9 @@ import (
 	"io"
 	"path/filepath"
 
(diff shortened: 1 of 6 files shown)


cmd/inpututil/input.go
```diff
@@ -4,9 +4,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"os"
 	"path/filepath"
 
+	"github.com/ubclaunchpad/inertia/local"
+
```
bobheadxi:
```suggestion
```

cmd/inpututil/input.go
```diff
@@ -26,13 +27,16 @@ func AddRemoteWalkthrough(
...
+		return err
+	}
 	sshDir := filepath.Join(homeEnvVar, ".ssh")
 	defaultSSHLoc := filepath.Join(sshDir, "id_rsa")
 
 	var response string
 	fmt.Println("Enter location of PEM file (leave blank to use '" + defaultSSHLoc + "'):")
-	_, err := fmt.Fscanln(in, &response)
+	_, err = fmt.Fscanln(in, &response)
 	if err != nil {
```
bobheadxi: while you're at it, could make this a one-liner? 😋 

```go
if _, err = fmt.Fscanln(in, &response); err != nil {
  ...
} 
```

daemon/inertiad/build/util.go
```diff
@@ -8,14 +8,17 @@ import (
...
 )
 
 // getTrueDirectory converts given filepath to host-based filepath if applicable
 // - Docker commands are sent to the mounted Docker socket and hence are
 // executed on the host, using the host's filepaths, which means Docker client
 // commands must use this function when dealing with paths
 func getTrueDirectory(path string) string {
-	return strings.Replace(path, "/app/host", os.Getenv("HOME"), 1)
+	homeDir, _ := local.GetHomePath()
+	return strings.Replace(path, "/app/host", homeDir, 1)
```
bobheadxi: we should remove this for now - *daemonside* windows compatibility is a much harder and nuanced issue, and a separate ticket is open for that (#407). Also trying to avoid introducing dependencies on `inertia/local` from the daemon for now, to avoid dependency creep 👌

yaoharry: Thanks for letting me know. I reverted the changes for util.go.

bobheadxi:
great work @yaoharry !

PR #577, opened 2019-02-21
client: configuration rework

### LINKED ISSUE #101: config: improve format of inertia configuration

bobheadxi:
:tickets: **Ticket(s)**: Closes #101

---

Many, many changes to configuration, which then caused many changes in `client`, which means a lot had to change in `cmd`, which means `local` must be updated, and I cleaned up a lot of misc stuff... therefore, huge mess of a PR.

Moral of the story: fix yo shit before building too much on top of your shit

### Rough tl;dr

* Inertia now configures **projects** and **remotes** separately
* "projects" are per-repo
    * stored in `/path/to/project/inertia.toml`
    * "profiles" provide build and deployment configuration, and can (or will be able to be) applied to a remote
    * new command chain `inertia project` to manage configuration (replaces `inertia config`)
* "remotes" are remote VPS instances, and are globally scoped
    * stored in `~/.inertia/inertia.global`
    * a remote has a set of "profiles", which maps projects to their applied profiles
    * differentiation between SSH-specific configuration and the rest of the stuff is improved (so that if you use Inertia's auth to access a daemon, Inertia wont be as annoying in pestering you about SSH stuff anymore)

### Misc changes

* shit tons of cleanup
* added some aliases to a few wordy commands

### inertia.toml

```toml
name = "test"
url = ""

[[profile]]
  name = "dev"
  branch = ""
  [profile.build]
    type = "dockerfile"
    buildfile = "Dockerfile.dev"

[[profile]]
  name = "staging"
  branch = ""
  [profile.build]
    type = "dockerfile"
    buildfile = "Dockerfile.staging"

```

### inertia.global

```toml
[[remote]]
  version = ""
  name = "dev"
  ip = "0.0.0.0"
  [remote.ssh]
    user = "bob"
    pemfile = ""
    ssh-port = ""
  [remote.daemon]
    port = "4043"
    token = ""
    webhook-secret = ""
    verify-ssl = false
  [remote.profiles]
    asdf = "asdf"
    oipo = "oiup"

[[remote]]
  version = ""
  name = "staging"
  ip = "0.0.0.0"
  [remote.ssh]
    user = "bob"
    pemfile = ""
    ssh-port = ""
  [remote.daemon]
    port = "4043"
    token = ""
    webhook-secret = ""
    verify-ssl = false
  [remote.profiles]
    fdsa = "fdsaf"
    wqrte = "erterh"

```
## Questions

- [ ] each remote definitely needs `version`, to track daemon version. do projects need it?
- [ ] is this scalable, potentially for things like Dockerhub deployment #201 ? 
- [ ] does it make sense?
- [x] given the new structure, is `toml` still the right choice? would `yml` be better? **toml**

nothing yet (tests are all busted)


bobheadxi:
paging @iKevinY , @rwblickhan , @brian-nguyen for thoughts 🙏

rwblickhan:
Does this mean we can deploy multiple projects to a single remote?

bobheadxi:
@rwblickhan no, but it does mean that you can more easily deploy a different project to the same remote to "recycle" them so to speak. For example:

```
inertia remote add local
inertia init
inertia project apply [profile-name] local # not implemented yet
```

then in another project:

```
inertia project apply [profile-name] local
```

then, when you run `inertia local up` in the two different projects, Inertia will apply your configured "profile" (but it'll still be one project per remote)

yaoharry:
I am for the shift to YAML because of the uniformity with cloud providers configuration files e.g gcloud and aws.

bobheadxi:
@yaoharry the thing is, `yml` was more designed to be handwritten, whereas `toml` is more consistent and generally generated - hence why I feel like it's been best for Inertia in the past. I think we'll probably stick with it for now, and hopefully we never get ot a point where you have to do a lot of hand-writing inertia configuration 😋 thanks for the feedback though!

@mRabitsky note taken about the nested properties - I've update the configuration to use lists instead of maps. We lose the uniqueness guarantees, but we have a much nicer-looking configuration now (let me know if this still isn't what you had in mind:

### inertia.toml

```toml
name = "test"
url = ""

[[profile]]
  name = "dev"
  branch = ""
  [profile.build]
    type = "dockerfile"
    buildfile = "Dockerfile.dev"

[[profile]]
  name = "staging"
  branch = ""
  [profile.build]
    type = "dockerfile"
    buildfile = "Dockerfile.staging"

```

### inertia.global

```toml
[[remote]]
  version = ""
  name = "dev"
  ip = ""
  [remote.profiles]
    asdf = "asdf"
    oipo = "oiup"

[[remote]]
  version = ""
  name = "staging"
  ip = ""
  [remote.profiles]
    fdsa = "fdsaf"
    wqrte = "erterh"

```

bobheadxi:
@mRabitsky re:

> As for the question of projects needing versions, perhaps that's something that Git's hashes can handle? (I think that's how it works in other similar systems).

By *version* I was referring to Inertia version. In `remote`, this is used to determine which [daemon build](https://cloud.docker.com/u/ubclaunchpad/repository/docker/ubclaunchpad/inertia/tags) to deploy - hence why I'm unsure if the version needs to be tracked elsewhere. Right now, if the format changes Inertia will just silently break where the format no longer matches

bobheadxi:
thanks @iKevinY 😋 i've updated the references

you don't have to read the code, sorry for the sinful PR 😂

PR #546, opened 2019-02-10
daemon: JSON responses for API

bobheadxi:
:tickets: **Ticket(s)**: Closes #502 

---

## :construction_worker: Changes

* [chi ](https://github.com/go-chi/chi) + middleware for routing needs
* basic JSON structure for as many things as I could update without demolishing things (package `api` and package `res`)
* update CLI and stuff as much as possible without radical changes

Here's the idea - all responses should be structured like this:

```go
// BaseResponse is the underlying response structure to all responses.
type BaseResponse struct {
	// Basic metadata
	HTTPStatusCode int    `json:"code"`
	RequestID      string `json:"request_id,omitempty"`

	// Message is included in all responses, and is a summary of the server's response
	Message string `json:"message"`

	// Error contains additional context in the event of an error
	Error string `json:"error,omitempty"`

	// Data contains information the server wants to return
	Data interface{} `json:"data,omitempty"`
}
```

`Data` is a json object of some sort. For example, `login` response looks like:

```js
{
   "code":200,
   "request_id":"bobbook/2Mch7LMzhj-000001",
   "message":"session created",
   "data":{
      "token":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzZXNzaW9uX2lkIjoiTEZ4cGVsMTc3R1hMd3d6Y281UjhNU0Z5eDZadWFYa0FQcFJpa25iRzhZTT0iLCJ1c2VyIjoiYm9iaGVhZHhpIiwiYWRtaW4iOmZhbHNlLCJleHBpcnkiOiIyMDE5LTAyLTEyVDA1OjAxOjM2LjgzMTQwNi0wODowMCJ9.0cszidOwITpIU3ZidENTwfnFBdorkkApXXfUkfUDsi0"
   }
}
```

and `totp/enable` looks like:

```js
{
   "code":200,
   "request_id":"bobbook/4NVeCCyrcK-000002",
   "message":"TOTP successfully enabled",
   "data":{
      "totp":{
         "secret":"D3YPABNVY6GBK3EH",
         "backup_codes":[
            "4aac6-44509",
            "a3415-2ccec",
            "5c14b-cbdb5",
            "3376e-0ca9c",
            "487ba-722f7",
            "63f2c-040c1",
            "9b0d6-4b561",
            "64442-157de",
            "497ee-07888",
            "be823-a7aa4"
         ]
      }
   }
}
```

[...]

PR #578, opened 2019-02-22
deploy: container history

seifghazi:
:tickets: **Ticket(s)**: for #293

---

Added containerBucket to manage container history for a deployed projects. 

Once a project is successfully deployed, the bucket is updated with the container ID and the git project's most recent commit hash. The commit hash can be later be used to identify successful 
builds and rollback to prior version when needed.

Before moving forward we might need to refactor data.go and extract env and container functionality to separate modules. 


daemon/inertiad/project/data.go
```diff
@@ -138,6 +139,14 @@ func (c *DeploymentDataManager) GetEnvVariables(decrypt bool) ([]string, error)
 	return envs, err
 }
 
+// AddBuiltContainer stores and tracks metadata from successful builds
+func (c *DeploymentDataManager) AddBuiltContainer(commitHash string, containerID string) error {
```
bobheadxi: for expandability sake I might make the value some kind of struct, so that we can add more fields to it in the future

for example:

```go
Container{ ID string }
```

and then we can have a signature like:

```suggestion
func (c *DeploymentDataManager) AddBuiltContainer(hash string, containers ...Container) error {
```

there `containers` utilizes the spread operator to allow an arbitrary number of `Container` (remember that docker-compose builds will have multiple containers, when we get around to supporting that)

that way, if we want to add build duration, you can update `Container`:

```go
Container{ ID string, BuildDuration int64 }
```

daemon/inertiad/project/deployment.go
```diff
@@ -314,6 +314,32 @@ func (d *Deployment) CompareRemotes(remoteURL string) error {
...
+func (d *Deployment) UpdateContainerHistory(cli *docker.Client) error {
+	// Retrieve container ID for recently deployed project
+	ctx := context.Background()
+	var containerID string
+	containers, err := cli.ContainerList(ctx, types.ContainerListOptions{})
+	if err != nil {
+		return err
+	}
+	for _, container := range containers {
+		if container.Names[0] == d.project {
```
bobheadxi: i think we might have to use `container.Names` to indicate build versions (I believe names must be unique, so if a project needs multiple containers for history, they cant all have the same name). We should add a new field to the container metadata (I think I sent you a link a while back) and put the project name there isntead, and check for it here

daemon/inertiad/daemon/up.go
```diff
@@ -97,5 +97,12 @@ func (s *Server) upHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+
+	// Update container management history
+	s.deployment.UpdateContainerHistory(s.docker)
+
+	logger.WriteSuccess("Project startup initiated!", http.StatusCreated)
```
bobheadxi: I don't think `logger` is used anymore, as of #546

daemon/inertiad/project/deployment.go
```diff
@@ -314,6 +314,32 @@ func (d *Deployment) CompareRemotes(remoteURL string) error {
...
+			containerID = container.ID
+		}
+	}
+	// Retrieve project hash
+	head, err := d.repo.Head()
+	if err != nil {
+		return err
+	}
+	// Update db with newly built container metadata
+	d.dataManager.AddBuiltContainer(head.Hash().String(), containerID)
```
bobheadxi: doesn't this return an error? we should probably check for it

daemon/inertiad/daemon/up.go
bobheadxi: this still stands 😋

daemon/inertiad/project/deployment.go
bobheadxi: jk this is fine for now I think

daemon/inertiad/project/data_test.go
```diff
@@ -65,36 +65,36 @@ func TestDataManager_EnvVariableOperations(t *testing.T) {
...
+			// Instantiate
+			c, err := NewDataManager(path.Join(dir, "deployment.db"), path.Join(dir, "key"))
+			assert.Nil(t, err)
 
-// 			// // Add
-// 			// err = c.AddBuiltContainer(tt.args.commitHash, tt.args.date)
-// 			// assert.Equal(t, tt.wantErr, (err != nil))
+			// Add
+			err = c.AddProjectBuildData(tt.args.projectName, tt.args.metadata)
+			assert.Equal(t, tt.wantErr, (err != nil))
```
seifghazi: Still need to add more tests before merging 🚀

daemon/inertiad/project/data.go
```diff
@@ -139,11 +140,46 @@ func (c *DeploymentDataManager) GetEnvVariables(decrypt bool) ([]string, error)
...
+			if err != nil {
+				return err
+			}
+
+			if err := projectBkt.Put([]byte(time.Now().String()), encodedMdata); err != nil {
+				return err
+			}
+		} else {
+			if err := c.UpdateProjectBuildData(projectName, encodedMdata); err != nil {
+				return err
```
bobheadxi: I would recommend adding more informative errors, for example:

```
return fmt.Errorf("oh no couldnt do x: %s", err.Error())
```

daemon/inertiad/project/data.go
```diff
@@ -138,6 +140,49 @@ func (c *DeploymentDataManager) GetEnvVariables(decrypt bool) ([]string, error)
 	return envs, err
 }
 
+// AddProjectBuildData stores and tracks metadata from successful builds
+// TODO: definitely need to add more context to errors
```
seifghazi: @bobheadxi 😁

bobheadxi: Ah hehe missed this

daemon/inertiad/project/data.go
```diff
@@ -138,6 +140,49 @@ func (c *DeploymentDataManager) GetEnvVariables(decrypt bool) ([]string, error)
...
+func (c *DeploymentDataManager) AddProjectBuildData(projectName string, mdata DeploymentMetadata) error {
```
seifghazi: @bobheadxi  you mind taking a look at the way I'm updating the db? Still getting a deadlock when adding two projects with the same name

seifghazi: Nvm 😂

daemon/inertiad/daemon/up.go
```diff
@@ -97,5 +97,12 @@ func (s *Server) upHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	// Update container management history following a successful build and deployment
+	err := s.deployment.UpdateContainerHistory(s.docker)
+	if err != nil {
+		stream.Error(res.ErrInternalServer("failed to update container history following build", err))
+	}
+
 	stream.Success(res.Msg("Project startup initiated!", http.StatusCreated))
+
```
bobheadxi: nitpick: extra space

daemon/inertiad/project/data.go
```diff
@@ -50,6 +52,7 @@ func NewDataManager(dbPath string, keyPath string) (*DeploymentDataManager, erro
 	}
 	if err = db.Update(func(tx *bolt.Tx) error {
 		_, err := tx.CreateBucketIfNotExists(envVariableBucket)
+		_, err = tx.CreateBucketIfNotExists(deployedProjectsBucket)
```
bobheadxi: we should check for the error on each bucket creation (right now the first one isn't being checked)

daemon/inertiad/project/data.go
```diff
@@ -138,6 +140,49 @@ func (c *DeploymentDataManager) GetEnvVariables(decrypt bool) ([]string, error)
...
+			if err := c.UpdateProjectBuildData(projectName, encodedMdata); err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+}
+
+// UpdateProjectBuildData updates existing project bkt with recent build's metadata
+func (c *DeploymentDataManager) UpdateProjectBuildData(projectName string, mdata []byte) error {
```
bobheadxi: this should either:

a) have the same signature as `AddProjectBuildData(projectName string, mdata DeploymentMetadata) error`
b) be removed, in favor of an option (`overwrite bool` for example) in `AddProjectBuildData`, or just always let `AddProjectBuildData` overwrite existing data

daemon/inertiad/project/data_test.go
```diff
@@ -65,6 +65,49 @@ func TestDataManager_EnvVariableOperations(t *testing.T) {
...
+		projectName string
+		metadata    DeploymentMetadata
+		numProjects int
+	}
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{"valid project build", args{"projectB", DeploymentMetadata{"hash", "ID", "status", "time"}, 2}, false},
```
bobheadxi: might want to add some potential error cases here - what if the project name is empty? etc.

daemon/inertiad/project/deployment.go
```diff
@@ -314,6 +325,59 @@ func (d *Deployment) CompareRemotes(remoteURL string) error {
...
+		ContainerStatus: containerStatus,
+		StartedAt:       containerStartedAtTime}
+
+	// Update db with newly built container metadata
+	err = d.dataManager.AddProjectBuildData(d.project, metadata)
+	if err != nil {
+		return fmt.Errorf("failure adding build metadata: %s", err.Error())
+	}
+
+	return err
```
bobheadxi: this can juust be `return nil`

PR #593, opened 2019-02-27
client: improved returns, debugging output

### LINKED ISSUE #236: client: improve client.Client function returns
Currently the vast majority of client.Client functions return `http.Reponse`, which is pretty unhelpful for the most part, and requires a lot of fiddling with response bodies etc on the user end

```go
// Reset shuts down deployment and deletes the contents of the deployment's
// project directory
func (c *Client) Reset() (*http.Response, error) {
	return c.post("/reset", nil)
}
```

We should update these so that they handle http responses internally (and return errors if needed) and instead of returning the raw responses, return some more predictable stuff. `Client.Reset()`, for example, could just be:

```go
func (c *Client) Reset() error {	
    /* handle http.Response and do stuff */
}
```

Theres a bit of code repetition in the client that we can probably improve on as well

This is not a particularly trivial PR, and might require a few passes.

bobheadxi:
:tickets: **Ticket(s)**: Closes #236

---

oops... this turned out bigger than expected 😓 

see #236 for context on this change

* client returns are no longer `http.Response`, but relevant structures instead
* to replace the loss of verbosity, a new `--debug` flag is now available to get debug output from the client
* client's user functions are now in `UserClient`, which you can get using `client.Client.GetUserClient()`
* improved use of `context.Context` for cleanliness of closure
* marginally improved tests

```
~/go/src/github.com/ubclaunchpad/inertia client/#236-client-rework*
❯ make cli

~/go/src/github.com/ubclaunchpad/inertia client/#236-client-rework*
❯ ./inertia local logs       
failed to connect to daemon: dial tcp 127.0.0.1:4303: connect: connection refused

~/go/src/github.com/ubclaunchpad/inertia client/#236-client-rework*
❯ ./inertia local logs --debug
DEBUG: request constructed: wss://127.0.0.1:4303/logs?container=%2Finertia-daemon&stream=true (authorized: false, verified: false)
failed to connect to daemon: dial tcp 127.0.0.1:4303: connect: connection refused
```


client/users_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+		// Check auth
+		assert.Equal(t, "Bearer "+fakeAuth, r.Header.Get("Authorization"))
+
+		render.Render(w, r, res.MsgOK("uwu"))
+	}))
+	defer testServer.Close()
+
+	var d = newMockClient(t, testServer).GetUserClient()
+	assert.NoError(t, d.RemoveUser(context.Background(), "yaoharry"))
```
yaoharry: uwu

PR #595, opened 2019-03-01
cmd: color and emoji

### LINKED ISSUE #147: cmd: colors + emoji
add some bling ✨

bobheadxi:
:tickets: **Ticket(s)**: Closes #147

---

* rename `output` to just `out`
* add `decor.go` with print wrappers and colour primitives (see `cmd/core/utils/out/decor.go`)
* prettified a few common commands (notably `init`, `${remote} init`, `remote add`, etc.)

## Todos

* [ ] is the `out` package ergonomic enough?
* [ ] hopefully merge https://github.com/kyokomi/emoji/pull/32, or stick to my fork of the emoji lib
* [x] add toggle to disable emoji and colours

```
make cli
./inertia
```

the output:

image


bobheadxi:
@yaoharry would appreciate if you could take a look! the colour library (https://github.com/fatih/color#color--) says that windows is supported, but who knows (its been archived for a little bit)

I also just added a new flag `--simple` to disable all decorations (color and emoji), and the `INERTIA_COLOR` env var can be used to disable colours

yaoharry:
> https://github.com/fatih/color#color--

Yeah looking at the library says it has Window support so thats nice. I think colors like cyan will default to blue on windows cmd, which imo is pretty ugly, but it's nice that we have a disable option.

bobheadxi:
> I think colors like cyan will default to blue on windows cmd, which imo is pretty ugly

Rip lol well we can keep an eye on feedback and maybe change the colour or disable colours on windows or improve the toggling (right now you can only toggle emoji AND colour at once via the flag, to just toggle one you have to set an end variable which is a bit less ergonomic)

Thanks for the reviews everyone!

Issue #607, opened 2019-03-17
config: allow inertia to not panic and kill everything on container death


bobheadxi:
As noted by @rwblickhan unknown containers running on the same VPS as inertia can cause unexpected behaviours, notably that Inertia usually cleans up everything if something fails.

I will be putting in a patch for `v0.5.x` but this will need to be forwardported for the significantly reworked configuration and client that has been done since `v0.5.2`. A forwardport should have a better implementation - maybe an array of `ignored_containers`

* [x] hotfix and release `v0.5.3` (PR: #608)
* [x] forwardport to `master` for `v0.6.x` releases (related: #626, pr: #649)
* [ ] come up with a longer-term solution to this maybe? (see: #633, https://github.com/ubclaunchpad/inertia/issues/633#issuecomment-546787504)

PR #608, opened 2019-03-17
config: hack in an ignore-death option

bobheadxi:
:tickets: **Ticket(s)**: for #607 

---

* in your config, manually add a field under remote configuration:

```toml
dont-kill-on-death = true
```

then `up`

more context in ticket. something like this will need to be separately implemented for what's on `master`

as above


rwblickhan:
Maybe it should be `no-kill-on-death`, that seems prettier 😛

